PR #12, opened 2016-06-03
Added StoredVerdict to verdicts

quoll:
This was originally done in the Verdict storage on CTIA, but has to move to CTIM


polygloton:
Related to threatgrid/ctia#277

Issue #25, opened 2016-06-14
Add event generators


polygloton:

gbuisson:
Events are not part of CTIM anymore

PR #26, opened 2016-06-14
Adding event generators, and fixing some issues along the way

polygloton:
Progress on #25


yogsototh:
LGTM!

src/ctim/generators/events.clj
```diff
@@ -0,0 +1,25 @@
+(ns ctim.generators.events
```
quoll: Should generators like this appear in a namespace that doesn't have "test" in the path?

polygloton: In the case of CTIM, yes, because it is a lib.  The other popular generators libs follow the same pattern.

PR #27, opened 2016-06-14
Refactor schema, using a BaseEntity

craigbro:
The major changes of this PR are:

1. All entities derive from  BaseEntity
2. We define SourcableObject and SourcedObject mixins for objects with optional, or required source/source_uri fields.
3. We define DescribableEntity for entities which can have a title, and descriptions.
4. Refactor code to use these entities

BaseEntity is defined as:

````
   :id ID
   :type s/Str
   :schema_version (describe s/Str "CTIM schema version for this entity")
   (s/optional-key :uri) URI
   (s/optional-key :revision) s/Int
   (s/optional-key :external_ids) [s/Str]
   (s/optional-key :timestamp) Time
   (s/optional-key :language) s/Str
   (s/optional-key :tlp) TLP
````

This add the following new keys:

* URI - optional full URI referring to this object
* revision - an interger 0-N that is iterated each time the object is edited
* timestamp - a timetap for when the object was last touched, used to resolve conflicts.  This is NOT the same as the created/modified fields in the Stored version o the entity, which are imlementation specific.
* external_ids - a list of opaque strings which are identifiers o this object in other systems.  Intended to be used by other system maintaining mirrors of their data in an instance of CTIA
* langage - a string identifier for what human language the text in this object is written in.  Default is english.
* tlp - Traffic Light protocol indicating the sharing rules for this data

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,23 +5,20 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.7.0"]
                  ;; what compojure-api 1.0.0 wants
-                 [prismatic/schema "1.0.4"]
+                 [prismatic/schema "1.1.1"]
                  ;; for describe
                  [metosin/ring-swagger "0.22.4"]
                  ;; for merge and such
-                 [metosin/schema-tools "0.7.0"]
+                 [metosin/schema-tools "0.9.0"]
                  ;; for generators
                  [org.clojure/test.check "0.9.0"]
                  [com.gfredericks/test.chuck "0.2.6"]
-                 [prismatic/schema-generators "0.1.0"]]
+                 [prismatic/schema-generators "0.1.0"
+                  :exclusions [prismatic/schema]]]
   :uberjar-name "ctim.jar"
   :resource-paths ["doc"]
   :profiles {:dev {:dependencies [[cheshire "5.5.0"]
-                                  [com.h2database/h2 "1.4.191"]
-                                  [org.clojure/test.check "0.9.0"]
-                                  [com.gfredericks/test.chuck "0.2.6"]
-                                  [prismatic/schema-generators "0.1.0"
-                                   :exclusions [prismatic/schema]]]
+                                  [com.h2database/h2 "1.4.191"]]
                    :resource-paths ["model"
                                     "test/resources"]}
              :test {:dependencies [[cheshire "5.5.0"]
--- a/src/ctim/schemas/campaign.clj
+++ b/src/ctim/schemas/campaign.clj
@@ -7,19 +7,26 @@
             [schema.core :as s]
             [schema-tools.core :as st]))
 
+(s/defschema TypeIdentifier
+  (s/enum "campaign"))
+
+
+
 (s/defschema Campaign
   "See http://stixproject.github.io/data-model/1.2/campaign/CampaignType/"
   (st/merge
-   c/GenericStixIdentifiers
-   {:valid_time (describe
+   c/BaseEntity
+   c/DescribableEntity
+   c/SourcableObject
+   {:type TypeIdentifier
+    :valid_time (describe
                  c/ValidTime
                  "timestamp for the definition of a specific version of a Campaign")
     ;; Extension fields:
-    :campaign_type  s/Str
-    :indicators rel/RelatedIndicators
-    :tlp c/TLP}
+    :campaign_type  s/Str}
    (st/optional-keys
     {:names (describe [s/Str] "Names used to identify this Campaign")
+     :indicators rel/RelatedIndicators
      :intended_effect (describe
                        [v/IntendedEffect]
                        (str "characterizes the intended effect of"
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: src/ctim/schemas/campaign.clj [hunk 1]
```diff
@@ -7,19 +7,26 @@
             [schema.core :as s]
             [schema-tools.core :as st]))
 
+(s/defschema TypeIdentifier
+  (s/enum "campaign"))
+
+
```
saintx: nit: whitespace

src/ctim/schemas/sighting.clj
```diff
@@ -16,22 +19,17 @@
...
+    :observed_time c/ObservedTime
+    :confidence v/HighMedLow
+    ;; how many times was it see
+    :count s/Int}
    (st/optional-keys
-    {:source s/Str
-     ;; if we have a source, we should have a source URI for more details
-     :source_uri c/URI
-     ;; The openC2 Actuator name that best fits the device that is
+    {;; The openC2 Actuator name that best fits the device that is
```
saintx: Should this say "openC2 Sensor" instead?

src/ctim/schemas/common.clj
```diff
@@ -37,6 +31,54 @@
...
+   })
+
+(s/defschema NewBaseEntity
+  "Base for New Entities, optionalizes ID and type and schema_version"
+  (st/merge
+   BaseEntity
+   (st/optional-keys
+    {:id ID
+     :type (describe s/Str "A valid entity type identifer")
+     :schema_version (describe s/Str "CTIM schema version for this entity")})))
```
yogsototh: Do we want the client to provide a schema version while creating a new entity?

In that case, does that mean that for CTIA we'll have to discard wrong version or keep a list of functions able to transform from one version to another one and make the schema upgrade automatic at creation/update?

src/ctim/schemas/common.clj
```diff
@@ -37,6 +31,54 @@
...
+
+(s/defschema SourcedObject
+  "An object that must have a source"
+  {:source s/Str
+   (s/optional-key :source_uri) URI})
+
+(s/defschema SourcableObject
+  "An object that MAY have a source"
+  {(s/optional-key :source) s/Str
+   (s/optional-key :source_uri) URI})
```
yogsototh: nit: maybe use `st/optional-keys`:

~~~clojure
(st/optional-keys
  {:source s/Str
   :source_uri URI})
~~~

src/ctim/schemas/common.clj
```diff
@@ -305,22 +348,22 @@
...
 
-(s/defschema Relation
+(s/defschema ObservableRelation
   {:id s/Num
    :timestamp Time
    :origin s/Str
    (s/optional-key :origin_uri) URI
-   :relation RelationType
+   :relation ObservableRelationType
    (s/optional-key :relation_info) {s/Keyword s/Any}
```
yogsototh: the `s/Any` here pose problems at schema and code generation. For example, the haskell client can't be generated correctly because of it and also, the testing generators can't use the :relation key because they cause coercion errors.

Could we use something like `s/Str` or `{s/Keyword s/Str}`

src/ctim/schemas/common.clj
gbuisson: I would make the accepted value to be an enum of the current schema version, this way the client would have to provide a version string matching the one currently used and fail otherwise.

project.clj
```diff
@@ -5,23 +5,20 @@
...
+                  :exclusions [prismatic/schema]]]
   :uberjar-name "ctim.jar"
   :resource-paths ["doc"]
   :profiles {:dev {:dependencies [[cheshire "5.5.0"]
-                                  [com.h2database/h2 "1.4.191"]
-                                  [org.clojure/test.check "0.9.0"]
-                                  [com.gfredericks/test.chuck "0.2.6"]
-                                  [prismatic/schema-generators "0.1.0"
-                                   :exclusions [prismatic/schema]]]
+                                  [com.h2database/h2 "1.4.191"]]
```
gbuisson: why do we use h2 on CTIM ?

src/ctim/schemas/common.clj
craigbro: We need to allow the client to specify that, and we need to accept all compatible versions -- and that should be determined via semantic versioning comparison of the schema versions.  AKA, if the first two numbers match.  This way, as we have different services generating data, we can know what needs transmogrifying (see Calvin and Hobbes) on import, or to clean up later.

Accepting the schema_version means we can set our own policies for wether we translate things on the fly, or flatly reject them if it's not a supported schema.

Since our ES schema is dynamic: strict -- we actually are tightly tied to the full version number, and are hence very sensitive to deviations in schema even if they are forward compatible (such as adding new fields).  I think we may want to change our dynamic mapping to be "allow but don't index" eventually.  I like keeping it strict now tho, since that helps us identify boogz.

src/ctim/schemas/common.clj
craigbro: Yes!  I think we should be doing that for sure.  Keys are always keywords or strings.  And we should define a schema that covers accepted values as well.

project.clj
craigbro: ask steve, I think his commit pulled that in 8^)

polygloton: It is an in memory SQL DB that is used in the tests

src/ctim/schemas/common.clj
polygloton: This is another great reason to get rid of the SQL store.

polygloton:
I would like to merge this and address comments in other issues.  Any objections?

yogsototh:
@polygloton I'm ok to merge it.

PR #31, opened 2016-07-11
schema generators now output entities with all their optional keys

### LINKED ISSUE #30: Generate full entity fixtures
related to #378

Currently our entity fixture generators only output the simplest scenario to match a schema, thus we can't simulate any of the optional keys.

Make it so that our entity schema generators output entities with most of their optional keys

gbuisson:
closes #30 
- use generators instead of complete to generate full entity fixtures
- all dates are now uniform: generated without formatting them as string
- schema_version on schemas is the current defined version
- added missing "type" optional keys

### DIFF
--- a/src/ctim/generators/schemas/actor_generators.clj
+++ b/src/ctim/generators/schemas/actor_generators.clj
@@ -1,5 +1,6 @@
 (ns ctim.generators.schemas.actor-generators
   (:require [clojure.test.check.generators :as gen]
+            [schema-generators.generators :as seg]
             [ctim.lib.time :as time]
             [ctim.schemas
              [actor :refer [NewActor StoredActor]]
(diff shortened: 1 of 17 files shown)


src/ctim/generators/schemas/actor_generators.clj
```diff
@@ -11,28 +12,20 @@
 
 (def gen-actor
   (gen/fmap
-   (fn [id]
-     (complete
-      StoredActor
-      {:id id}))
-   (gen-id/gen-short-id-of-type :actor)))
+   (fn [[s id]]
+     (into s {:id id}))
```
yogsototh: When keys are known `assoc` is faster than `into`. This remark in not a blocker of course, I generally forgot this optimisation myself most of the time.

~~~clojure
user=> (time (doseq [v (range 10000)] (assoc {:a "a"} :id v)))
"Elapsed time: 1.261498 msecs"
nil
user=> (time (doseq [v (range 10000)] (into {:a "a"} {:id v})))
"Elapsed time: 6.176702 msecs"
~~~

yogsototh:
LGTM!

PR #32, opened 2016-07-11
issue-178: OpenC2 commands for network and SDN actuator types

jyoverma:
closes https://github.com/threatgrid/ctia/issues/178


src/ctim/schemas/coa.clj
```diff
@@ -9,6 +9,11 @@
 (s/defschema TypeIdentifier
   (s/enum "coa"))
 
+(s/defschema StructuredCOA
+  {;; :id and :idref must be implemented exclusively
+   :id c/ID})
```
saintx: Can this be gotten by merging with c/BaseEntity?

jyoverma:
Thanks for the feedback Alex. StructuredCOA is a field of COA and so I thought the schema  belonged here but if you think it should be under BaseEntity, I can move it there. 
Please advice,
Jyoti

saintx:
No, I don't think moving the StructuredCOA *into* the BaseEntity would be a good idea.   Rather, StructuredEntity could *extend* the BaseEntity schema.

Since the BaseEntity has an :id key already, it might be useful to extend that, rather than make a new entity type from scratch that overlaps that schema.  There are also some other useful parts that CTIA wants to see, like the :type and :tlp keys. Here's the whole list:

``` clj
(s/defschema BaseEntity
  {;; :id and :idref must be implemented exclusively
   :id ID
   :type s/Str
   :schema_version (describe (s/enum ctim-schema-version) "CTIM schema version for this entity")
   (s/optional-key :uri) URI
   (s/optional-key :revision) s/Int
   (s/optional-key :external_ids) [s/Str]
   (s/optional-key :timestamp) Time
   (s/optional-key :language) s/Str
(s/optional-key :tlp) TLP})
```
using this as a base, your StructuredCOA schema could look like so:

``` clj
(s/defschema StructuredCOA
  c/BaseEntity)
```

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
+(ns ctia.schemas.openc2-network
+  (:require [schema.core :as s]
+            [ctia.schemas.openc2vocabularies :as openc2v]))
```
saintx: this will need to be ctim.schemas.openc2vocabularies

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
+(ns ctia.schemas.openc2-network
```
saintx: s/ctia/ctim

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
...
+  (:require [schema.core :as s]
+            [ctia.schemas.openc2vocabularies :as openc2v]))
+
+
+
+
+(s/defschema BGPBlackhole
+  {:type (s/enum "BGPBlackhole")
+   :host s/Str
+   })
```
saintx: In before @AlexBaranosky tells you not to dangle your closing braces and parens :)

jyoverma:
Yes, that makes sense. I'll make the changes.

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
...
+  (s/enum "ALERT" 
+          "DROP"
+          "DENY"
+          "LOG"
+          "PASS"
+          "REJECT"))
+
+(s/defschema TrafficType
+  {:protocol Protocol
+   :sourceAddress s/Str
```
saintx: I think as a matter of general principle, key names should not be camel case.  In CTIM, our keys are intended to be written to JSON, and so we use the JSON style with the underscore, like so:

``` clj
:source_address s/Str
```

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
...
+(s/defschema Honeypot
+  {:type (s/enum "Honeypot")
+   :permissibleIps [s/Str]
+   :acl NetworkACL 
+   :routes {:prefix s/Str
+            :NextHop s/Str
+            :NextHopType s/Str}
+   })
+
+(def EncapsulationType
```
saintx: Should this be defschema?

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
...
+  (s/enum "GRE"
+          "VXLAN"))
+
+(s/defschema BlockModifierType
```
saintx: I think we can dispense with the use of the term "Type" throughout all of our schema names--we did so through the rest of the CTIM, so this would just be:

``` clj
(defschema BlockModifier
  ...
)
```

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,117 @@
...
+   :permissibleIps [s/Str]
+   :acl NetworkACL 
+   })
+
+(s/defschema Honeypot
+  {:type (s/enum "Honeypot")
+   :permissibleIps [s/Str]
+   :acl NetworkACL 
+   :routes {:prefix s/Str
+            :NextHop s/Str
```
saintx: What's the difference between the next hop and the next hop type?

src/ctim/schemas/openc2_network.clj
AlexBaranosky: No dangling allowed!!

jyoverma:
Alex, I've made the changes that you suggested. Please review. Thanks!

src/ctim/schemas/openc2_network.clj
```diff
@@ -0,0 +1,104 @@
...
+            #(= "NetworkACL" (:type %)) NetworkACL
+            #(= "BGPBlackhole" (:type %)) BGPBlackhole
+            #(= "DNSSinkhole" (:type %)) DNSSinkhole)})
+
+
+  
+
+(s/defschema ContainModifier
+  {:type (s/eq "Contain")
+   :method (s/conditional
```
gbuisson: Can we use something else than conditional here?
It doesn't play well with swagger.

jyoverma: Sure .... I'm new to Clojure so didn't know any better. Please recommend an alternative approach.

Thanks,
Jyoti

polygloton: I have to disagree with the request to avoid conditional.  As far as I can tell, the only problem is with the swagger UI.  The JSON schema that gets generated is correct.  We shouldn't have to write bad code because of a weakness in the swagger UI.  If it is really a bother, then just fix the bug, please.

gbuisson: The problem is not with swagger UI, but the swagger spec itself, this is not a bug, it's a choice from the developers on the spec.

Violating the spec not only breaks Swagger UI, but also code generation and makes our API incompatible with any standard Swagger tool.

there are many huge threads about this issue, mainly:
https://github.com/OAI/OpenAPI-Specification/issues/57

I was suggesting this change to know if there is any "smart decision" we could make on this schema to avoid writing a workaround, for instance isn't @craigbro previously proposed solution eligible here? instead of having:

```clojure
:method (s/conditional
             #(= "NetworkACL" (:type %)) NetworkACL
             #(= "BGPBlackhole" (:type %)) BGPBlackhole
             #(= "DNSSinkhole" (:type %)) DNSSinkhole)})
```

maybe we could use something like this:

```clojure
(st/optional-keys 
{:method_NetworkACL NetworkACL
 :method_BGPBlackhole BGPBlackhole
 :method_DNSSinkhole DNSSinkhole})
```

yogsototh: There are other issues relatives to the use of conditionals:

- They break automatic generative testing, even if I believe it is not too difficult to fix manually
- It make ES mapping difficult (maybe even impossible), and I imagine it would also make difficult to make SQL tables if one day we want to use SQL.

src/ctim/schemas/coa.clj
```diff
@@ -2,13 +2,43 @@
...
             [schema.core :as s]
             [ring.swagger.schema :refer [describe]]
             [schema-tools.core :as st]))
 
 (s/defschema TypeIdentifier
   (s/enum "coa"))
 
+(s/defschema StructuredCOA
+  (st/merge
+   c/BaseEntity))
```
craigbro: Since this is not a top level entity, it should not be based on BaseEntity.

src/ctim/schemas/coa.clj
```diff
@@ -39,8 +69,8 @@
                          " one or more related courses of action"))
      ;; Not provided: handling
      ;; Not provided: parameter_observables ;; Technical params using the CybOX language
-     ;; Not provided: structured_COA ;; actionable structured representation for automation
-     })))
+     :structured_coa OpenC2COA})))
```
craigbro: So, like the conditional problem below, we need to handle different types of structured COAs.

StructuredCOA should have a type value, and then a key based on that type.  So `{:structured_coa_type "openc2" :open_c2_coa {...}}`

This way it can have multiple types of structures/specification.

Also, on the whole FOO_type vs "type" I am partial to the :foo_type - mostly because in ES, we could not have different mappings for :type fields in different locations in the object, they would all have to be string/enum.  That's my only argument.  I do think it introduces some verbosity tho, and makes writing a generic structured accessor harder.

src/ctim/schemas/openc2_network.clj
polygloton: I still recoil at modifying schemas to work around incidental complexity in Swagger.  In OAI/OpenAPI-Specification#57 they acknowledge that they may fix this in a future release.  Lack of "oneOf" (eg "conditional") is, at best, a feature that swagger doesn't support yet, at worst, a bad engineering decision.

In response to Yann's bullets:,

- I don't see a ton of value in automatic data generation based on schemas alone since it ignores the semantics of your data and it gives a false sense of confidence that you are testing the right thing.  
- I don't believe that it makes mapping impossible.  You can create multiple tables/indexes, one for each type (there is a small set of possible schemas), and do a join based on ID and type.

We should not be complicating @jyoverma's life any further with this argument.  :-)  @craigbro came up with a good plan that we should follow:

> [11:37 AM] Craig Brozefsky: my evil plan would be

> [11:37 AM] Craig Brozefsky: we just have a dummied schema for swagger
> [11:37 AM] Craig Brozefsky: but our stored schema is fully conditionalized and all that
> [11:38 AM] Craig Brozefsky: so we get all the prismatic specificity, and then if we fix the tooling, we drop the dummy

jyoverma:
I've incorporated the feedback requests in the latest checkin.

thanks,
Jyoti

PR #38, opened 2016-07-15
issue-36: Add schema for Kill Chain

### LINKED ISSUE #36: Add cyber kill chain to TTP schema
Add cyber kill chain to TTP schema.

saintx:
Adds schema for Kill Chain and Kill Chain Phases.

closes #36

### DIFF
--- a/src/ctim/schemas/ttp.clj
+++ b/src/ctim/schemas/ttp.clj
@@ -106,12 +106,11 @@
                     (str
                      "specifies other TTPs asserted to be related to this cyber"
                      " threat TTP"))
-     :source (describe s/Str "source of this cyber threat TTP")})
+     :source (describe s/Str "source of this cyber threat TTP")
+     :kill_chains [v/KillChain]})
    {;; Extension fields:
     :ttp_type (describe s/Str "type of this TTP")
     :indicators (describe rel/RelatedIndicators "related indicators")
-    ;; Not provided: kill_chain_phases
-    ;; Not provided: kill_chains
     ;; Not provided: handling
     ;; Not provided: related_packages (deprecated)
     }))
--- a/src/ctim/schemas/vocabularies.clj
+++ b/src/ctim/schemas/vocabularies.clj
@@ -190,6 +190,15 @@
           "Traffic Diversion"
           "Unauthorized Access"))
 
+(s/defschema KillChain
+  (s/enum "Reconnaissance"
+          "Weaponization"
+          "Delivery"
+          "Exploitation"
+          "Installation"
+          "Command & Control"
+          "Actions on Objectives"))
+
 (s/defschema LocationClass
   (s/enum "Internally-Located"
           "Externally-Located"
--- a/src/ctim/schemas/common.clj
+++ b/src/ctim/schemas/common.clj
@@ -182,7 +182,7 @@
   {:start_time
    (describe Time
              "Time of the observation.  If the observation was made over a period of time, than this ield indicated the start of that period")
-   
+
    (s/optional-key :end_time)
    (describe Time
              "If the observation was made over a period of time, than this field indicates the end of that period")})


gbuisson:
LGTM

src/ctim/schemas/ttp.clj
```diff
@@ -106,12 +106,11 @@
                     (str
                      "specifies other TTPs asserted to be related to this cyber"
                      " threat TTP"))
-     :source (describe s/Str "source of this cyber threat TTP")})
+     :source (describe s/Str "source of this cyber threat TTP")
+     :kill_chains [v/KillChain]})
```
sashank: This Kill chain phase should be associated with Indicators will not be appropriate at TTP level

src/ctim/schemas/vocabularies.clj
```diff
@@ -190,6 +190,15 @@
           "Traffic Diversion"
           "Unauthorized Access"))
 
+(s/defschema KillChain
```
sashank: LockHeed Martin's phases of kill chain are from attacker's perspective. CTIA Modelling is from defender's perspective the subtle nuance is that , few phases may not be applicable and could cause confusion. For example, "Weaponization" is an activity done by attacker (a verb) and may not have concrete "indicator" (a noun) to describe this activity. Also the difference between "exploitation" and "installation" is very subtle in reality it may not really help the defender to differentiate between them. So would recommend dropping "Weaponization" and "Installation" stages. From a defenders perspective less number of states eases his investigation.

deandebeer: While I agree with you about attacker versus defender, LM defined this from the perspective of a defender tracking an attacker's actions. Today all IR/Intel analysts are aware of, and leverage, the kill chain in this manner. Removing fields will only serve to add confusion. Re: Exploitation & Installation. Exploitation refers to the triggering of a vuln, the presence of in-memory or disk artifacts validating the triggering of an exploit. This can also mean the exploitation of a user via email or drive-by. Installation, in the Kill Chain, refers to the installation, via dropper or weaponized document, of the implant or second stage binary/malware. This is commonly used for the downloading of attacker toolkits for remote access or lateral movement.

sashank: Hmm makes sense. We can leave these standard steps as is.

src/ctim/schemas/ttp.clj
craigbro: While Indicators can refer to KC phases, we also want them to be associated with TTPs.  For example when we identify a network of C&C servers, we would want to associate a KC phase with that of "Command & Control"  just like we may associate a set of servers known to serve up exploits...

If you look at how TG uses Indicators, usually identifying specific malware famlies, you'll also see that we can have ALOT of different indicators for all of the variations of the family.  We want to put the KC phase on the TTP, so that analysts don't have to remember that when they write the next Indicator for next weeks variant.  They can just say, "X Exploit kit" ... or "Zeus" and get the relevant KC phase associated with it.

src/ctim/schemas/vocabularies.clj
craigbro: We have BIs that can distinguish between Exploitation and Installation.

src/ctim/schemas/ttp.clj
sashank: It might get tricky. In some cases, the set of servers (identified by IPs)  of same malware family might be used/mapped to different phases of KC. for example, few ip addresses in same set could be CnC and few could be used to upload exfil data (Action phase ) , few could serve initial malware payloads (delivery). So assigning a collection of IPaddresses (indicators) to particular phase of KC. So ideally one or more KC phases are associated with a TTP.  (Similar to STIX - [TTP](http://stixproject.github.io/data-model/1.2/ttp/TTPType/) abstraction ) .  So in current modelling if it is conveying the plurality of distinct KC phases per TTP then we are good.

PR #39, opened 2016-07-26
Converted the project to cljc

### LINKED ISSUE #37: Convert CTIM to cljc
Convert the whole project to cljc so it can be used from clojurescript

gbuisson:
closes #37 

- Tests are also converted run them with either 
`lein doo node node once`or `lein doo phantomjs test once`
-  A quick cljs sample will be added later


yogsototh:
LGTM if you squash all commits!

src/ctim/generators/id.cljc
```diff
@@ -17,18 +19,27 @@
...
 (def gen-path
-  (chuck/string-from-regex
-   #"(\/([\w][-\w]{0,5}(\.[-\w]{1,4}){0,3})){1,3}"))
+  #?(:clj (chuck/string-from-regex
+           #"(\/([\w][-\w]{0,5}(\.[-\w]{1,4}){0,3})){1,3}")
+     :cljs
+     (gen/fmap (fn [[f s]]
+                 (str "/" f "/" s))
+               (gen/tuple (gen-str-3+ gen-char-alpha-lower)
+                          (gen-str-3+ gen-char-alpha-lower)))))
```
polygloton: I think you should just use the :cljs implementation for both here.  Maybe we can get rid of chuck.

src/ctim/lib/time.cljc
```diff
@@ -41,21 +51,15 @@
...
   (time/after? (coerce-to-datetime left)
                (coerce-to-datetime right)))
 
-(defn sql-now []
-  (coerce-to-sql-time (now)))
-
-(def to-sql-time coerce-to-sql-time)
-
-(def from-sql-time coerce-to-date)
-
```
polygloton: Why are these deleted?  Are they not used?

gbuisson: Yes since we removed sql support there is no need for this anymore.
also, as suggested previously we should split this code into a common toolset lib shared by our projects.

PR #41, opened 2016-08-05
Add Relationship entity

### LINKED ISSUE #40: Implement Relationship entity
Need to allow people to make relations between objects (source, relation, destination)

Should contain:
- Source: URI to origin object
- Relation: a new vocabulary
- Destination: The subject object

polygloton:
Entity to relate other entities to each other.

Closes #40

### DIFF
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -396,6 +396,8 @@
 
 (defn stored-schema
   "Given a schema X generate a StoredX schema"
-  [type-name a-schema]
-  (st/merge a-schema
-            BaseStoredEntity))
+  ([_ a-schema] ;; deprecated
+   (stored-schema a-schema))
+  ([a-schema]
+   (st/merge a-schema
+             BaseStoredEntity)))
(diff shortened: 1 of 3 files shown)


src/ctim/generators/schemas/judgement_generators.cljc
```diff
@@ -1,7 +1,7 @@
 (ns ctim.generators.schemas.judgement-generators
   (:require [clj-momo.lib.time :as time]
             [clojure.test.check.generators :as gen]
-            [ctim.schemas.common :as schemas-common]
+            [ctim.schemas.shared.common :as schemas-common]
```
craigbro: why rename?

polygloton: Do you mean why move common under shared?  I had moved relationships to shared.relationships to avoid confusion with similarly named relationship entity.  When I did that, I asked myself, "What else is shared?" and moved those things as well.

craigbro: I would prefer to not add another layer of namespace hierarchy.

And instead change the name of the other ObservedRelationship or perhaps move to something like ObservedData from STIX

polygloton:
We want to specify a default vocabulary, but keep the :relationship_type field open (allow strings).  The work to be done in #60 will allow this.

craigbro:
Let's please get this merged with just "String" as the type, and later we can make it into an "open-vocabulary" when Flanders is operational

polygloton:
Will do.

PR #56, opened 2016-09-08
Fixed ID URL regex so that it recognizes entity types containing hyphens

### LINKED ISSUE #55: ID regex fails in some cases
Matching URL IDs with hyphens fails.

For example, in ctim.domain.id:

```clojure
(re-matches long-id-re "http://localhost:3001/ctia/exploit-target/exploit-target-d51dfc7b-df40-46a4-9b06-c396e3dfdbcf")
```

Results in nil.

polygloton:
Closes #55

### DIFF
--- a/src/ctim/domain/id.cljc
+++ b/src/ctim/domain/id.cljc
@@ -6,7 +6,7 @@
   "(([a-z][-a-z]+)-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
 
 (def url-pattern
-  "(https?):\\/\\/([-\\da-zA-Z][-\\da-zA-Z.]*)(:(\\d+))?((\\/[-\\w.]+)*)\\/ctia\\/([a-z]+)\\/")
+  "(https?):\\/\\/([-\\da-zA-Z][-\\da-zA-Z.]*)(:(\\d+))?((\\/[-\\w.]+)*)\\/ctia\\/([a-z][-a-z]+)\\/")
 
 (def short-id-re (re-pattern short-id-pattern))
 (def url-re (re-pattern url-pattern))
--- a/test/ctim/domain/id_test.cljc
+++ b/test/ctim/domain/id_test.cljc
@@ -1,8 +1,8 @@
 (ns ctim.domain.id-test
   (:require [ctim.domain.id :as id]
             [ctim.generators.id :as gen]
-            #?(:clj [clojure.test :refer [deftest is]]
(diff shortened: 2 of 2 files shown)


src/ctim/domain/id.cljc
```diff
@@ -6,7 +6,7 @@
   "(([a-z][-a-z]+)-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
 
 (def url-pattern
-  "(https?):\\/\\/([-\\da-zA-Z][-\\da-zA-Z.]*)(:(\\d+))?((\\/[-\\w.]+)*)\\/ctia\\/([a-z]+)\\/")
+  "(https?):\\/\\/([-\\da-zA-Z][-\\da-zA-Z.]*)(:(\\d+))?((\\/[-\\w.]+)*)\\/ctia\\/([a-z][-a-z]+)\\/")
```
quoll: Is there a reason why IDs can't start with a dash?

test/ctim/domain/id_test.cljc
```diff
@@ -85,3 +85,15 @@
...
+  (testing "with hyphens in the entity type"
+    (is (= #ctim.domain.id.CtiaId{:hostname "localhost",
+                                  :short-id "exploit-target-d51dfc7b-df40-46a4-9b06-c396e3dfdbcf",
+                                  :path-prefix nil,
+                                  :port 3001,
+                                  :protocol "http",
+                                  :type "exploit-target"}
+           (id/long-id->id
+            (str "http://localhost:3001/ctia/exploit-target/"
+                 "exploit-target-d51dfc7b-df40-46a4-9b06-c396e3dfdbcf"))))))
```
quoll: It's only a test, so I'm less worried, but I would extract the short-id portion into a def:
```
(def test-id "exploit-target-d51dfc7b-df40-46a4-9b06-c396e3dfdbcf")

(deftest test-long-id->id
  (testing "with hyphens in the entity type"
    (is (= #ctim.domain.id.CtiaId{:hostname "localhost",
                                  :short-id test-id,
                                  :path-prefix nil,
                                  :port 3001,
                                  :protocol "http",
                                  :type "exploit-target"}
           (id/long-id->id
            (str "http://localhost:3001/ctia/exploit-target/" test-id))))))
```

polygloton:
fixed :)

src/ctim/domain/id.cljc
polygloton: I think you mean entity (eg models).  That would just be weird.

PR #61, opened 2016-09-23
Use sets for Bundle containers, support for non CTIM data

gbuisson:
- Use sets instead of vector for entity containers, this eases entity deduping
- We need Bundles to support non CTIM schemas added :other and :other_refs in this regard
- Added a custom set generator like vector to handle max complexity

### DIFF
--- a/src/ctim/schemas/bundle.cljc
+++ b/src/ctim/schemas/bundle.cljc
@@ -47,7 +47,7 @@
     :sighting_refs ReferenceSet
     :ttp_refs ReferenceSet
     :verdict_refs ReferenceSet
-    :other_refs #{s/Any}}))
+    :other_refs ReferenceSet}))
 
 (s/defschema Bundle
   (st/merge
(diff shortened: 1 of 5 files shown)


src/ctim/schemas/bundle.cljc
```diff
@@ -17,35 +17,37 @@
...
+    :coa_refs ReferenceSet
+    :exploit-target_refs ReferenceSet
+    :feedback_refs ReferenceSet
+    :incident_refs ReferenceSet
+    :indicator_refs ReferenceSet
+    :judgement_refs ReferenceSet
+    :sighting_refs ReferenceSet
+    :ttp_refs ReferenceSet
+    :verdict_refs ReferenceSet
+    :other_refs ReferenceSet}))
```
craigbro: `other_refs` should be `datatable_refs` and `datatables` for embeedded objects.

gbuisson:
I will close this one and open another for the datatable refs as we're no longer planning to use sets for now.

PR #62, opened 2016-10-02
Using flanders to define entity schemas and documentation

### LINKED ISSUE #60: Use flanders to define our entities
Based on the [spike-schema-tree](https://github.com/polygloton/ctim/tree/spike-schema-tree) branch.

- [ ] Generate plumatic schemas on-demand
- [ ] Generate model documentation

polygloton:
Closes #60


src/flanders/core.cljc
```diff
@@ -0,0 +1,176 @@
...
+;; Defining Branch Nodes
+;; ----------------------------------------------------------------------
+
+(declare key)
+(declare enum)
+(declare seq-of)
+
+(defn entry [key_ type_ & {:keys [description reference required?]
+                         :or {required? true}}]
+  (let [key_ (if (keyword? key_) (key key_) key_)
```
yogsototh: Just a note to remark it took me long time to understand it wasn't `clojure.core/key` even if the declare is just above.

polygloton: I was trying to build a terse DSL here, but might have gone too far :)

polygloton: Good point

polygloton: I just didn't want to shadow 'key, so 'k would have been better

src/ctim/generators/schemas/actor_generators.cljc
```diff
@@ -2,29 +2,31 @@
...
-     (cond-> (dissoc s :id :valid_time)
-       id (assoc :id id)
-       start-time (assoc-in [:valid_time :start_time] start-time)
-       end-time (assoc-in [:valid_time :end_time] end-time)))
-   (gen/tuple
-    (seg/generator NewActor)
-    (maybe (gen-id/gen-short-id-of-type :actor))
-    ;; complete doesn't seem to generate :valid_time values, so do it manually
-    common/gen-valid-time-tuple)))
+  (let [NewActor (fs/get-schema csa/NewActor)]
```
gbuisson: sounds like you don't use it, perhaps you should remove that def or move it at the root of the ns.

src/ctim/schemas/bundle.cljc
```diff
@@ -11,60 +11,67 @@
...
-    :sightings [StoredSighting]
-    :ttps [StoredTTP]
-    :verdicts [StoredVerdict]}))
+(def objects-entries
+  (f/optional-entries
+   (f/entry :actors [StoredActor])
+   (f/entry :campaigns [StoredCampaign])
+   (f/entry :coas [StoredCOA])
+   (f/entry :exploit-targets [StoredExploitTarget])
+   (f/entry :feedbacks [StoredFeedback])
```
gbuisson: it would be nice if we could avoid repeating `f/entry`, maybe we could do something with the wrapping `optional-entries` ?

polygloton: (f/etnry ...) could be replaced with [...], like

```clojure
[:coas [StoredCoa]
  :comment "this is a stored COA"]
```

or we could just alias f/entry with 'e as in

```clojure
(e :coas [StoredCoA]
 :comment "This is a stored COA")
```

It all sucks, just in different ways.

src/ctim/generators/schemas/actor_generators.cljc
polygloton: Good catch, will fix.

Issue #65, opened 2016-10-06
Improve the generated markdown


polygloton:
Iterate on improving the markdown that is generated by the flanders type definitions

- [x] Nested maps should be documented un-nested with links in the top level map
- [x] References should be markdown links, not just URIs
- [x] All maps should have a name, and most should have descriptions
- [x] Any URL in a description should be a markdown link
- [x] Add :comments to the nodes (developer notes in the documentation)
- [x] Truncate long schemas, like "(enum a b c ...", which can be done since the values are listed

Also do some simple refactoring

- [ ] Standardize map definitions (syntax evolved during conversion of CTIM)
  - [x] Use [Type] in entries instead of (seq-of Type)
  - [x] Use map/map-of/def-map-type consistently
  - [x] Use #{Type ...} in entries instead of (enum #{Type ...})
  - [x] def-entity-type should allow references to be defined

polygloton:
- [ ] Add Usage Notes to leaf nodes

polygloton:
- [ ] There should be an index page that points to the various generated documents
- [ ] Need to delete old schema docs in CTIM and CTIA (threatgrid/ctia#423)

craigbro:
I think you should just make :description be markdown, that way if I want a link, I can write it out as markdown, and I get all the other markdown behavior.

yogsototh:
I believe we hit the issue for :description used in swagger (which I don't think support markdown) and adding another field, :long-description that would be able to handle markdown and be used for more in depth documentation.

craigbro:
Since markdown is just text, why is the lack of support for it in swagger an isue?

yogsototh:
In swagger-ui the text describing a field is intended be "short".

For example this is what swagger-ui display when an user click on Model description:

~~~
Log {
tags (Array[string], optional): An optional list of strings ,
_event_type (string, optional): The type of event for example syslog, cisco_syslog, wsa_log ,
_format (string, optional): The optional logstash codec, default is plain another common is json ,
@timestamp (string): Event occured at ,
_collector (string): The name of the collector this event came through ,
_raw (string): The raw, original event text ,
@version (string): Event schema version ,
_source (string): The source, a file path, socket or url path or other identifier specific to _input ,
_encoding (string, optional): The character set of the event, default UTF-8 ,
_input (string): The name of the input module this event came via
}
~~~

If we write a long description, it will certainly be hard to read in swagger-ui.

But we also sometime need to be able to add a long, in-depth description.
[...]

PR #68, opened 2016-10-07
Better generated documentation

polygloton:
Part of #65

### DIFF
--- a/doc/actor.md
+++ b/doc/actor.md
@@ -4,7 +4,7 @@
 http://stixproject.github.io/data-model/1.2/ta/ThreatActorType/
 
 
-## MapEntry: :id -> String
+## MapEntry :id ∷ String
 
 * This entry is required
 
@@ -15,7 +15,7 @@ http://stixproject.github.io/data-model/1.2/ta/ThreatActorType/
   * The URI of this entity.
   * Plumatic Schema: Str
 
-## MapEntry: :type -> String
+## MapEntry :type ∷ String
 
 * This entry is required
 
@@ -25,7 +25,7 @@ http://stixproject.github.io/data-model/1.2/ta/ThreatActorType/
 * String Value
   * Plumatic Schema: Str
 
-## MapEntry: :schema_version -> String
+## MapEntry :schema_version ∷ String
 
 CTIM schema version for this entity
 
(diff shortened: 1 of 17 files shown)


doc/actor.md
```diff
@@ -7,570 +8,521 @@ http://stixproject.github.io/data-model/1.2/ta/ThreatActorType/
...
-### String Value
-
-The URI of this entity.
-
-* Plumatic Schema: Str
+* String Value
+  * The URI of this entity.
+  * Plumatic Schema: Str
 
 ## MapEntry: :type -> String
```
yogsototh: I'm not fond of using "->" here at is somehow used for functions.

https://github.com/polygloton/ctim/blob/c9e8e932cea244abd9505f9773f6317dac09d3b7/src/flanders/markdown.cljc#L64

As we somehow anotate some "type" I would rather prefer to use `:` or maybe `::`.

src/flanders/markdown.cljc
```diff
@@ -51,38 +67,48 @@
...
+    (str "  * Plumatic Schema: "
+         (if (fp/sequential? loc)
+           (str "[" schema "]")
+           schema)
+         "\n")))
 
 (defn- ->values [{v :values}]
   (when (and v (> (count v) 1))
-    (str "* Allowed Values: " (sort (seq v)) "\n")))
+    (str "  * Allowed Values: " (sort (seq v)) "\n")))
```
yogsototh: I find this hard to read. Even if it makes more line, I would prefer to have a list of items here.
I don't think the depth is kept (even if it should certainly be 3).

If there are case were it might not be 3 we could use `html` `<ul>` and `<li>` in markdown.

If the depth is 3 that should do it:

~~~clojure
(->> (sort (seq v))
       (map #(str "      * " % "\n"))
~~~

polygloton: I was thinking about that too.  :)

doc/actor.md
polygloton: I think the ":"'s make it confusing though.  It triggers some sort of tag matching on github, too.  I was thinking about ruby's hashmap literals, which in retrospect is "=>" and not "->".  What do you think about using "=>"?

yogsototh: Yes I believe that if you could "protect" the keyword it would look better. But the current version looks good to me.

PR #74, opened 2016-10-10
Improving generated docs

### LINKED ISSUE #65: Improve the generated markdown

polygloton:
Closes #65


polygloton:
The docs are generated and are included in this commit.  Also the top level README has instructions.

PR #79, opened 2016-10-12
Flanders schema gen reimplementation

### LINKED ISSUE #72: Support reference sets
In flanders, support #{Str} type schemas

### LINKED ISSUE #78: Flanders should generate schemas for any node type
Currently it only generates them for maps

polygloton:
Fixes it so that any node can be turned into a schema, properly supports conditional, and adds sets as value types.

Closes #78 and closes #72

### DIFF
--- a/src/ctim/generators/schemas/indicator_generators.cljc
+++ b/src/ctim/generators/schemas/indicator_generators.cljc
@@ -17,8 +17,7 @@
   (gen/fmap
    (fn [[s id]]
      (assoc s :id id))
-   (gen/tuple (seg/generator (fs/get-schema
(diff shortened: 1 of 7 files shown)


polygloton:
The implementation of flanders.schema/->schema-tree is a lot better.  It really is extensible now.  It is a lot more flexible.  The only drawback is that it now uses recursion (and therefore the stack).  That shouldn't be a problem in practice.

polygloton:
Note that EitherType nodes are no longer replaced with AnyType nodes by default.  There is a fn for that now, which we should use for the swagger APIs in CTIA.

PR #80, opened 2016-10-12
CTIA Flanders support

gbuisson:
Those are small fixes I made while upgrading CTIA to latest 0.2.0-SNAPSHOT


polygloton:
Need to update the docs with "lein doc" after changing schemas

polygloton:
LGTM (I will update the docs)

Issue #82, opened 2016-10-13
Resolve "relationship" NS name conflicts


polygloton:
- [ ] relationships.cljc should be references.cljc
- [ ] observed_relationship.cljc should be for observable relationships
- [ ] the entity ObservedRelationship should just be Relationship in relationship.cljc

PR #88, opened 2016-10-14
Relationship reorg

craigbro:
Rename ObservedRelationship to Relationship

Also put the ctia.schema.relationships/* schema into that
ctima.schema.relationship namepace so all schema relatied to
relationships between domain objects are in the same namespace.


polygloton:
Related to #82

Issue #83, opened 2016-10-13
Documentation improvements


polygloton:
Iterate on the CTIM documentation

polygloton:
- [ ] Table of contents for each entity, ideally a table
- [ ] Return links should show the name of the thing being returned to
- [ ] More better descriptions

PR #94, opened 2016-10-17
Documentation JSON examples and updated graph

polygloton:
Related to #83


yogsototh:
That's great!

LGTM!

If I understand correctly the flanders code. For now you cannot override the default value by type. For example it would have been great to be able to precise a default `id` string for each object, as they all follow the same pattern `<entity name>-UID`.

polygloton:
@yogsototh You are right, that is a feature that I plan on adding.  This PR is just a fast hack.

PR #103, opened 2016-11-09
Can set :id when creating events

polygloton:
In support of threatgrid/ctia#442

### DIFF
--- a/test/ctim/schemas/campaign_test.cljc
+++ b/test/ctim/schemas/campaign_test.cljc
@@ -59,8 +59,8 @@
                                   :source "source"
                                   :relationship "rel"}]
           :confidence "High"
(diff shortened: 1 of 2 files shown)


test/ctim/schemas/campaign_test.cljc
```diff
@@ -59,8 +59,8 @@
                                   :source "source"
                                   :relationship "rel"}]
           :confidence "High"
-          :activity {:date_time #inst "2016-02-11T00:40:48.212-00:00"
-                     :description "activity"}})))
+          :activity [{:date_time #inst "2016-02-11T00:40:48.212-00:00"
```
yogsototh: I've fixed this in #106 and it is unrelated to the issue.
So it should be cleaner to simply remove these modification along this issue. Furthermore #106 as an updated doc.

PR #112, opened 2016-11-16
Judgement :severity is changed to a HighMedLow

### LINKED ISSUE #102: Severity should be v/HighMedLow instead of a number
Numbers are so vacuous, and present a false specificity.  We should have High Medium Low severity, and we should provide some strong examples and explanations of what they are.

polygloton:
Closes #102


polygloton:
Replaced by #116

PR #119, opened 2016-11-21
Schema changes, to become rel 0.4.0

polygloton:
This will merge multiple schema changes that we want to release together.


polygloton:
Need to get CTIA ready for CTIM 0.4.0 before merging this into master.

PR #124, opened 2016-12-06
Defining and using spec generators

### LINKED ISSUE #58: Generated entities need to build domain aware values
We want to return an error when references are not URLs in CTIA.  Tests that use generated entities (that contain references) get random strings in the reference and ID fields.  The current generators are not specific enough to create sufficiently accurate examples of our domain objects.

In other words, don't just take a schema and generate samples.

Related to threatgrid/ctia#212

### LINKED ISSUE #123: Bundle should use references instead of IDs
Currently the bundle entity uses "short IDs" as the values in its *_ref fields.  Need to use full URL references instead.

polygloton:
- Removes all of the old schema generators
 - Adds properties on DDL leaf nodes for defining generators
 - Also using references in Bundle (issue #123)

Closes #123, closes #58 and supports threatgrid/ctia#419.


polygloton:
This allow us to specify :gen along with :spec in our DDL, which will allow us to validate (and generatively test) our entities in the API.  Most of the entity generating code is removed from CTIM.  I will modify the CTIA generative code to use the generators provided by clojure.spec.  While working on this code I found that our old generators were not generating "complete" examples like we thought.  Using this code (as seen in the tests in the PR), we can solve that problem.  However, this makes generating things like Bundles super slow.  I will delete/replace any generators in CTIA that are too slow (rather than waste more time trying to solve that problem) and use examples as well.

src/ctim/schemas/common.cljc
```diff
@@ -2,30 +2,51 @@
...
+         :spec (cs/and string? :ctim.domain.id/long-id)
+         :gen gen-id/gen-url-id))
 
 (defn ref
   "Make a custom Reference"
   [& opts]
   (apply (partial assoc Reference) opts))
 
+(defn ref-for-type
+  [type & opts]
```
polygloton: Need to do something with opts

PR #134, opened 2017-02-21
Fixing verdict schemas

### LINKED ISSUE #133: Verdict and StoredVerdict should follow the standard pattern
Most entities have an :id and :schema-version fields, which are mixed in with base-entity-entries.  Verdict is an exception to this, but it shouldn't be.  It should be treated like any other entity.  Instead, StoredVerdict adds those fields.  This causes problems in CTIA because we want to be able to convert a StoredVerdict into a Verdict by stripping off standard store keys.

Make Verdict and StoredVerdict conform to the common pattern.

polygloton:
Closes #133

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,7 +1,7 @@
 (def schema-tools-version "0.9.0")
 (def schema-version "1.1.3")
 
-(defproject threatgrid/ctim "0.4.3"
+(defproject threatgrid/ctim "0.4.4-SNAPSHOT"
   :description "Cisco Threat Intelligence Model"
   :url "http://github.com/threatbrain/ctim"
   :license {:name "Eclipse Public License"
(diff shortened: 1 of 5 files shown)


project.clj
```diff
@@ -1,7 +1,7 @@
 (def schema-tools-version "0.9.0")
 (def schema-version "1.1.3")
 
-(defproject threatgrid/ctim "0.4.3"
+(defproject threatgrid/ctim "0.4.4-SNAPSHOT"
```
quoll: Is this the correct PR to be updating the version to SNAPSHOT? I presume that the version update is just piggy-backing on this PR

polygloton: The change will require a new version.  The alternative is to bump the version directly in master, which I'm fine doing, but I'm also fine doing it in the PR.

PR #139, opened 2017-03-08
Adding disposition sorting logic

polygloton:
Also cleaned up a test

### DIFF
--- a/src/ctim/domain/disposition.cljc
+++ b/src/ctim/domain/disposition.cljc
@@ -1,0 +1,18 @@
+(ns ctim.domain.disposition)
+
+(def ^:private weights
+  {"Clean"      1
+   "Malicious"  2
+   "Suspicious" 3
+   "Unknown"    4})
+
+(defn importance [a b]
+  (compare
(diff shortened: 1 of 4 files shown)


src/ctim/domain/disposition.cljc
```diff
@@ -0,0 +1,18 @@
+(ns ctim.domain.disposition)
+
+(def ^:private weights
+  {"Clean"      1
+   "Malicious"  2
+   "Suspicious" 3
+   "Unknown"    4})
```
yogsototh: I'm not 100% certain this has some importance, but I prefer to make the remark.

Regarding the disposition in verdict we also have "Common" with value 4 and "Unknown" for value 5:

~~~clojure
                  "clean" 1
                  "malicious" 2
                  "suspicious" 3
                  "common" 4
                  "unknown" 5
~~~

polygloton: This is only ranking the importance, so I guess the assumption is that common is the same as default, which is 99.  If that is wrong, then I should change it.

PR #157, opened 2017-06-02
Fixing bug where UtcDateTime comparisons failed in CLJS

### LINKED ISSUE #156: CLJS sorting error: Can't compare DateTimes
When sorting judgements in CLJS, it fails with an error (if the judgements use DateTimes instead of date strings).  UtcDateTime objects are not comparable.

polygloton:
Closes #156

* Relies on latest clj-momo.lib.clj-time.core which makes UtcDateTime comparable.
* Also adds some guarding coercions that make sure input dates are acceptable.

### DIFF
--- a/src/ctim/domain/sorting.cljc
+++ b/src/ctim/domain/sorting.cljc
@@ -1,6 +1,7 @@
 (ns ctim.domain.sorting
   (:require
    [clj-momo.lib.clj-time.core :as time]
+   [clj-momo.lib.clj-time.coerce :as time-coerce]
    [ctim.domain.validity :as validity]))
 
 ;;------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/ctim/domain/sorting.cljc
```diff
@@ -15,22 +16,30 @@
...
+                         (validity/valid-now? dt-now j1_start j1_end))]
     (if (not= 0 result1)
       result1
       (let [result2 (compare j1_disp j2_disp)]
         (if (not= 0 result2)
           result2
           ;; Start-time is reverse sorted
-          (compare j2_start j1_start))))))
+          (compare (time-coerce/to-long j2_start)
+                   (time-coerce/to-long j1_start)))))))
```
yogsototh: The compare function is a bit hard to follow for me.
Would it be possible to rename some temporary values with more meaningful names?

Instead of `result1`, `result2` why not `valid-time-comparison`, `disposition-comparison`?

yogsototh: Wouldn't `time/after?` be faster than first translating the date to long then comparing the time? 

Also, there might be direct lower level `.compare` method to be used if the internal representation is some joda time DateTime. I'm not sure about that at all for cljs.

src/ctim/domain/sorting.cljc
```diff
@@ -15,22 +16,30 @@
...
-                         (validity/valid-now? dt-now j1))]
+    :as _judgement-2_}]
+  (let [j1_start (time-coerce/to-internal-date j1_start)
+        j1_end   (time-coerce/to-internal-date j1_end)
+        j2_start (time-coerce/to-internal-date j2_start)
+        j2_end   (time-coerce/to-internal-date j2_end)
+
+        ;; Validity is reverse sorted so that true comes before false
+        result1 (compare (validity/valid-now? dt-now j2_start j2_end)
+                         (validity/valid-now? dt-now j1_start j1_end))]
```
yogsototh: `valid-now?` also call `to-internal-date` on its arguments. I had to take a look in `clj-momo` to check that it is indeed an identity function when called on the right argument.

Even though it should cost almost nothing, couldn't we get rid of the overriding named function.

Typically the `j1_start` in the let override the `j1_start` of the function. I believe we should get rid of the `let` here and use the `jX_{start,end}` of from the function arguments directly.

src/ctim/domain/sorting.cljc
polygloton: I'll play around with it.

polygloton: I can make UtcDateTime satisfy IComparable.

src/ctim/domain/sorting.cljc
polygloton: I'm fond of this simple pattern of re-binding, but I'll change it to make you happy.  :-)

And yes, when the judgement inputs are proper CTIM entities (meaning they use Date/UtcDateTime objects), then the coercion is basically identity.  When it is anything else, there is a good chance that it will get coerced correctly.  Of course there is a performance cost then.  There doesn't seem to be any better way to handle the situation, except maybe throw errors.

polygloton: The reason that it doesn't use fn args directly is because the start date-time objects may get used twice, and I want to avoid coercing them twice.  We can use the end-times directly though.

polygloton:
I'm going to implement sorting in ES, probably with [script based sorting(https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#_script_based_sorting).  However, this is a good fix that will resolve the bug now.

PR #159, opened 2017-06-22
Add target object to Sighting entity

msprunck:
Connected to threatgrid/iroh#518

### DIFF
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -335,7 +335,10 @@
     "email"
     "imei"
     "imsi"
-    "amp-device"})
+    "amp-device"
+    "amp_computer_guid"
+    "hostname"
+    "mac_address"})
 
 (def-enum-type ObservableTypeIdentifier
   observable-type-identifier
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -338,7 +338,9 @@
     "amp-device"
     "amp_computer_guid"
     "hostname"
-    "mac_address"})
+    "mac_address"
+    "file_name"
+    "file_path"})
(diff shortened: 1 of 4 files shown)


craigbro:
Thanks for putting this together, it captures most of the information we need, and shows the most direct encoding of it.

We should be using existing CTIM patterns where possible.  So target should have a set of observables which define, or "triangulate" it.  So, encode the IPs, guid, and mac addr as observables.  We also need to handle targets that may have multiple IP interfaces.

I also think we should be using the taxonomy of devices that we use in sighting.sensor field to identify the target.  Some possible target types are servers, specific web services, email accounts, network devices (routers,firewalls), mobile devices (endpoint still?)

/code
:target {:type "endpoint"
              :observables
                {:type "hostname" :value "Demo_Cta"}
                {:type "amp_computer_guid" :value "68e94bf7-e239-4821-90d6-b7eaa0233443"}
                {:type "ip" :value "100.213.110.122"}
                {:type "ip" :value "136.184.130.98"}
                {:type "mac_address" :value "85:28:cb:6a:21:41"}}
              :properties_data_table
                {.. data table object encoding operating system, version etc... }}

msprunck:
The distinction between internal and external IP address is not important ?

src/ctim/schemas/vocabularies.cljc
```diff
@@ -335,7 +335,12 @@
     "email"
     "imei"
     "imsi"
-    "amp-device"})
+    "amp-device"
```
yogsototh: using a dash is not coherent with the new observable types that use an underline char.

msprunck: I use the underscore for new values only. For backward compatibility, I kept the dash for old values.

The dash character is not valid for GraphQL enums

PR #160, opened 2017-07-26
Adding length limits to string fields

polygloton:
Part of threatgrid/ctia#578


test/ctim/test_helpers/core.clj
```diff
@@ -3,7 +3,8 @@
             [clojure.spec.test :as stest]
             [clojure.test.check.generators :as gen]
             [ctim.generators.common :as cgc]
-            [flanders.spec :as fs]))
+            [flanders.spec :as fs]
+            [schema.core :as schema]))
```
polygloton: These requires can be deleted

polygloton:
These changes seem to noticeably slow down the generative tests.  Not sure why.

test/ctim/specs/fields_test.clj
```diff
@@ -0,0 +1,753 @@
...
+          true  (rand-str 5000)  new-kw    new-ex
+          true  (rand-str 5000)  stored-kw stored-ex
+          false (rand-str 5001)  plain-kw  plain-ex
+          false (rand-str 5001)  new-kw    new-ex
+          false (rand-str 5001)  stored-kw stored-ex
+          false (rand-str 10000) plain-kw  plain-ex
+          false (rand-str 10000) new-kw    new-ex
+          false (rand-str 10000) stored-kw stored-ex))))
+
+(deftest test-field-validators
```
msprunck: What is the purpose of this test ? 
To validate ShortString, MedString, LongString and Markdown schemas or to check that all string fields are correctly specified in entity schemas.

polygloton: Its meant to assert that the fields have the correct constraints.

polygloton:
I think the dev should have to change a test if they change a field.  I tried to make the tests as concise as possible, so hopefully they would only have to change the name of one fn call.

PR #167, opened 2017-09-04
Observable types cleaning

msprunck:
- Removing `amp_device` observable type. `amp_computer_guid` is used for AMP connector GUID.
- Replacing all '-' characters by '_'


msprunck:
@craigbro Should we remove the `device` observable type previously used for mac address ?

msprunck:
Covered by #191

PR #170, opened 2017-09-18
Adding some Integer constraints

polygloton:
Testing that :revision, and :count on Sighting, is a positive integer and is between 0 and 1,000,000, inclusively.

### DIFF
--- a/src/ctim/lib/generators.clj
+++ b/src/ctim/lib/generators.clj
@@ -3,6 +3,10 @@
    [clojure.string :as str]
    [clojure.test.check.generators :as gen]))
 
+(defn pos-int-max [max-num]
+  (gen/such-that #(<= % 0)
+                 gen/pos-int))
+
 (defn string-max-len [max-len]
   (gen/fmap (fn [base-str]
               (when base-str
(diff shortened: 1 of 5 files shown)


polygloton:
I need to change the way that dates are generated to guarantee that end_time is never less than start_time.  Going to get some dinner and think it over.

test/ctim/specs/sighting_observed_time_test.clj
```diff
@@ -0,0 +1,39 @@
...
+            [ctim.schemas.sighting :as si]
+            [ctim.test-helpers.core :as th]
+            [ctim.examples.sightings :as e]
+            [flanders.spec :as fs]))
+
+(use-fixtures :once
+  th/fixture-spec-validation
+  (th/fixture-spec si/Sighting "test.sighting"))
+
+(deftest test-sighting-observed-time-spec-validation
```
msprunck: Could you add a test to check that the observed time is valid when `start_time` is equal to `end_time`?

polygloton: Test added :)

Issue #171, opened 2017-09-29
E1331: Add Attack Pattern, Malware and Tool to CTIM


msprunck:
Epic threatgrid/iroh#1331

Add the following entities from [STIX 2.0](https://oasis-open.github.io/cti-documentation/stix/intro) to CTIM:
- [Attack Pattern](https://docs.google.com/document/d/1IvkLxg_tCnICsatu2lyxKmWmh1gY2h8HUNssKIE-UIA/edit#heading=h.axjijf603msy)
- [Malware](https://docs.google.com/document/d/1IvkLxg_tCnICsatu2lyxKmWmh1gY2h8HUNssKIE-UIA/edit#heading=h.s5l7katgbp09)
- [Tool](https://docs.google.com/document/d/1IvkLxg_tCnICsatu2lyxKmWmh1gY2h8HUNssKIE-UIA/edit#heading=h.z4voa9ndw8v)

See [Cyber Threat Intelligence Repository of ATT&CK](https://github.com/mitre/cti) for examples.

PR #172, opened 2017-10-03
Adding Attack Pattern, Malware and Tool entities from Stix 2.0

msprunck:
Connected to #171 

In order to import objects from the [Cyber Threat Intelligence Repository of ATT&CK catalog expressed in STIX 2.0], we created these new entities:

- [x] Attack Pattern
- [x] Malware
- [x] Tool

An `:external_references` property has been added to all CTIM entities. External references are used to describe pointers to information represented outside of CTIM (See [External References](https://docs.google.com/document/d/1dIrh1Lp3KAjEMm8o2VzAmuV0Peu-jt9aAh1IHrjAroM/pub#h.72bcfr3t79jx) in STIX 2.0).

Ex:
```edn
{:external_references
 [{:source_name "veris"
   :external_id "0001AA7F-C601-424A-B2B8-BE6C9F5164E7"
   :url
   "https://github.com/vz-risk/VCDB/blob/master/data/json/0001AA7F-C601-424A-B2B8-BE6C9F5164E7.json"
   :hashes
   {:SHA-256
    "6db12788c37247f2316052e142f42f4b259d6561751e5f401a1ae2a6df9c674b"}}]}
```

The old TTP object has been removed with all of its dependencies (value objects, vocabularies)

Documentation update will be subject to a new PR


saintx:
This is awesome.

craigbro:
According to saintx, this is awesome.  I concur, even if he is afraid to say it in text.

src/ctim/schemas/common.cljc
```diff
@@ -350,6 +364,17 @@
   (vals disposition-map)
   :description "String verdict identifiers")
 
+(def-map-type KillChainPhase
```
saintx: Why are we not actually defining the kill chain phases as vocabulary here?  They would be the same as those defined in https://en.wikipedia.org/wiki/Kill_chain

- Reconnaissance: Intruder selects target, researches it, and attempts to identify vulnerabilities in the target network.
- Weaponization: Intruder creates remote access malware weapon, such as a virus or worm, tailored to one or more vulnerabilities.
- Delivery: Intruder transmits weapon to target (e.g., via e-mail attachments, websites or USB drives)
- Exploitation: Malware weapon's program code triggers, which takes action on target network to exploit vulnerability.
- Installation: Malware weapon installs access point (e.g., "backdoor") usable by intruder.
- Command and Control: Malware enables intruder to have "hands on the keyboard" persistent access to target network.
- Actions on Objective: Intruder takes action to achieve their goals, such as data exfiltration, data destruction, or encryption for ransom.

msprunck: STIX 2.0 does not link any vocabulary to the [Kill Chain Phase](https://docs.google.com/document/d/1dIrh1Lp3KAjEMm8o2VzAmuV0Peu-jt9aAh1IHrjAroM/edit#heading=h.i4tjv75ce50h). I can add the suggested vocabulary in CTIM.

msprunck: The ATT&CK catalog uses ATT&CK tactics for STIX `kill_chain_phases.phase_name`. As explained in the [ATT&CK introduction and overview](https://attack.mitre.org/wiki/Introduction_and_Overview), the ten tactic categories were derived from the later stages (control, maintain, and execute) of a seven-stage Cyber Attack Lifecycle.

Image of Yaktocat

Tactics:

- persistence
- privilege-escalation
- defense-evasion 
- credential-access
- discovery
- lateral-movement
- execution
- collection
- exfiltration
- command-and-control

What vocabulary should we use?

msprunck: In the STIX specification, the Kill Chain Phase has two fields:
- kill_chain_name (ex: mitre-attack, lockheed-martin-cyber-kill-chain)
- phase_name (ex: reconnaissance or weaponization if the kill_chain_name is lockheed-martin-cyber-kill-chain)

The phase_name vocabulary depends on the kill_chain_name.

Should we restrict the kill_chain_name to `lockheed-martin-cyber-kill-chain`?

polygloton:
Should an external reference just be a new type of relationship?

craigbro:
I don't think it should be a new type of Relationship for the following reasons:

1. It would deviate from stix 2.0 model, which has cognitive/interop costs
2. It's not between two CTIM entities, and would make navigating relations difficult.

src/ctim/schemas/common.cljc
```diff
@@ -117,6 +118,25 @@
...
+         :gen #?(:clj gen/open-vocab-chars
+                 :cljs nil)))
+
+(defn open-vocab
+  "Defines an OpenVocab type with a suggested vocabulary"
+  [values]
+  (assoc OpenVocab
+         :open? true
+         :values values))
+
```
polygloton: I like `OpenVocab` and `open-vocab`, good idea :)

PR #191, opened 2018-02-09
Simplify Incident Model and add Severity to Indicator

craigbro:
Simplified the Incident Model, removing the parts of the model that we
will not be using for the first phase implementation of the
shared Incident Response Service

Added Severity to Indicator, since it is important information to have
for pattern Indicators that have no Judgements, and also it described
the Indicator better.


yogsototh:
As this is a breaking change I think the schema_version and ctia version should be changed accordingly.

PR #192, opened 2018-02-13
Added attack pattern abstraction levels

quoll:

PR #193, opened 2018-02-14
Issue-190: Add attack pattern abstraction levels

### LINKED ISSUE #190: Add abstraction-level to Attack Patterns.
We've recently been trying to find an idiomatic way to represent categories of attack patterns in CTIM.  There's a useful concept from the CAPEC data that we might consider adding to our model for attack patterns, which I think would solve this problem, and which would simplify and possibly even accelerate mapping between the indicators from AMP & TG and our various attack patterns.

All attack patterns in CAPEC have an "abstraction" field, which is one of the values in ["Category" "Meta" "Standard" "Detailed"].  This seems like a very useful bit of data, and think we could benefit from adding "abstraction-level" to our attack-pattern object, and defining a vocabulary for it.

There are some attack patterns, such as ["Privilege Abuse"](https://capec.mitre.org/data/definitions/122.html) which are meta attack patterns.  We might also use the meta abstraction level for things like "Trojan", "Virus", and other high level common concepts that apply to thousands of more specific techniques.  By contrast, ["XML Entity Blow-Up"](https://capec.mitre.org/data/definitions/201.html) is a much more specifically detailed attack pattern, which is a very specific subtype of a type of Privilege Abuse.

We might have a record of a malware family that uses "privilege abuse", but not that specific "XML Entity Blow-Up" pattern.  Or, we could have a malware object that indicates both.

Attack Patterns from the MITRE ATT&CK data set each have a category on the "ATT&CK Matrix", and it would be trivial to represent that relationship using this mechanism as well.

Finally, this use of abstraction levels for attack patterns would allow RET to start at a high level of abstraction for its `indicator->attack-pattern` mappings, and even add their own higher level abstractions, and then return later on to add more specific `standard` and `detailed` attack pattern entries when more information becomes available.

## Proposal
1. Add an optional value to the [attack-pattern schema](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/attack_pattern.cljc) `:abstraction-level`, which refers to a Vocabulary.
2. Define a new Vocabulary for abstraction levels, including `["Category" "Meta" "Standard" "Detailed"]
3. No new relationship types should be required.

saintx:
This adds support for an optional `abstraction_level` entry in attack patterns,
which can be used to define attack patten categories, meta attack patterns,
aggregate attack patterns, and detailed attack patterns in addition to the
standard ones found in data sets like MITRE CAPEC & ATT&CK.

Closes #190

### DIFF
--- a/src/ctim/examples/attack_patterns.cljc
+++ b/src/ctim/examples/attack_patterns.cljc
@@ -21,6 +21,7 @@
    :source_uri "http://example.com"
    :name "Bootkit"
    :description "A bootkit is a malware variant that modifies the boot sectors of a hard drive"
+   :abstraction_level "standard"
    :kill_chain_phases [{:kill_chain_name "mitre-attack"
                         :phase_name "persistence"}]
    :x_mitre_data_sources ["API monitoring" "MBR" "VBR"]
--- a/src/ctim/schemas/attack_pattern.cljc
+++ b/src/ctim/schemas/attack_pattern.cljc
@@ -1,7 +1,8 @@
 (ns ctim.schemas.attack-pattern
   (:require #?(:clj  [flanders.core :as f :refer [def-entity-type def-map-type def-eq]]
                :cljs [flanders.core :as f :refer-macros [def-entity-type def-map-type def-eq]])
-            [ctim.schemas.common :as c]))
+            [ctim.schemas.common :as c]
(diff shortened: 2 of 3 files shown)


saintx:
Replaces https://github.com/threatgrid/ctim/pull/192

PR #200, opened 2018-02-21
Sighting model improvements

gbuisson:
- HighMedLow should have a new value, Info - which indicates it's purely informational.

- add `internal` at the root of the model `s/Bool`

- add `severity` at the root of the model, using HighMedLowInfo Vocabulary

- add `resolution` at the root of the model `s/Str` or `s/enum`, this field indicates if the sensor that is reporting the `Sighting` already took action on it, for instance a Firewall blocking the IP.
  - detected
  - blocked
  - allowed
  - contained

- add `observed_time` to `Sighting/Target`, make it required
- make `Sighting/Target` a collection for th Sighting summary feature

### DIFF
--- a/src/ctim/schemas/sighting.cljc
+++ b/src/ctim/schemas/sighting.cljc
@@ -1,15 +1,28 @@
 (ns ctim.schemas.sighting
-  (:require [ctim.schemas.common :as c]
-            [ctim.schemas.relationship :as rel]
-            [ctim.schemas.vocabularies :as v]
-            #?(:clj  [flanders.core :as f :refer [def-entity-type def-eq def-map-type]]
-               :cljs [flanders.core :as f :refer-macros [def-entity-type def-eq def-map-type]])))
+  #?@
+  (:clj
+   [(:require
+     [ctim.schemas.common :as c]
+     [ctim.schemas.relationship :as rel]
+     [ctim.schemas.vocabularies :as v]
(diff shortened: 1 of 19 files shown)


msprunck:
Could you also convert the target field to a collection?

src/ctim/schemas/sighting.cljc
```diff
@@ -1,15 +1,28 @@
...
-            [ctim.schemas.relationship :as rel]
-            [ctim.schemas.vocabularies :as v]
-            #?(:clj  [flanders.core :as f :refer [def-entity-type def-eq def-map-type]]
-               :cljs [flanders.core :as f :refer-macros [def-entity-type def-eq def-map-type]])))
+  #?@
+  (:clj
+   [(:require
+     [ctim.schemas.common :as c]
+     [ctim.schemas.relationship :as rel]
+     [ctim.schemas.vocabularies :as v]
```
msprunck: Why do you put these namespaces in the the reader conditionals part? It's common to clj and cljs

gbuisson: cider and clj-format do that when you automatically reformat the ns,
I wonder if I prefer manually maintain the conditionals or let the tools do it.

msprunck:
Don't forget to update `ctim-schema-version` when you will release the next version of CTIM

Issue #198, opened 2018-02-15
TTP Model Refinements


saintx:
These model refinements are needed to support the addition of CAPEC data to AMP Global Intel, and to develop mappings between the TG indicators and TTP data from Mitre and elsewhere.

- [x] [#186 Make attack pattern, malware, and tool sourcable.](https://github.com/threatgrid/ctim/issues/186)
- [x] [#190 Add abstraction level to Attack Patterns](https://github.com/threatgrid/ctim/issues/190)
- [x] [#196 Add Weakness entity](https://github.com/threatgrid/ctim/issues/196)
- [x] [#195 Add Vulnerability entity](https://github.com/threatgrid/ctim/issues/195)
- [x] [#197 Malware abstraction level](https://github.com/threatgrid/ctim/issues/197)
- [x] [#194 Remove ExploitTarget](https://github.com/threatgrid/ctim/issues/194)

- [x] CTIA Support for the above model changes
- [x] Data Migration in INT, TEST, and PROD

PR #206, opened 2018-04-26
Add the Weakness Entity

### LINKED ISSUE #196: Add Weakness entity
Add a Weakness entity which we can use to represent Weaknesses, such as those enumerated in https://cwe.mitre.org

- [x] Extends `common/base-entity-entries`
- [x] Weakness should be `sourceable` and `describable`.
- [x] Add an example for the entity, if none exists.

## Example Weakness
- https://cwe.mitre.org/data/definitions/95.html

msprunck:
Epic #198

Closes #196 

Based on the CWE XSD http://cwe.mitre.org/data/xsd/cwe_schema_v6.0.1.xsd

I've not added these fields:

- Extended_Description
- Related_Weaknesses (Relationship)
- Weakness_Ordinalities
- Demonstrative_Examples
- Observed_Examples
- Taxonomy_Mappings
- Related_Attack_Patterns
- References (external_references)
- Content_History
- Name
- Status

### DIFF
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -439,6 +439,8 @@
   malware-abstraction-labels
   :description "Malware Abstraction level")
 
+;; ---- Weakness CWE
+
 (def software-phases
   #{"Policy"
     "Requirements"
@@ -583,7 +585,8 @@
     "Moderate"
     "SOAR Partial"
     "Opportunistic"
-    "Limited"})
+    "Limited"
+    "None"})
 
 (def-enum-type DetectionEffectiveness
   detection-effectivness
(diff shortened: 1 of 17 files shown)


msprunck:
@saintx Could you check the content fits your needs? If it's ok, I will add examples and tests

src/ctim/schemas/vocabularies.cljc
```diff
@@ -490,3 +490,255 @@
...
+    "Protection Mechanism"
+    "Session Management"
+    "Signals"
+    "String Processing"
+    "Functional-Area-Independent"})
+
+(def-enum-type FunctionalArea
+  functional-areas
+  :open? true
+  :description "functional area of software appear"
```
saintx: "Defines the different functional areas of software in which the weakness may appear." is a better description.

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,202 @@
...
+                              "in which the weakness is most likely to occur"))
+   (f/entry :affected_resources (f/set-of v/SystemResource)
+            :description (str "identify system resources that can be affected "
+                              "by an exploit of this weakness"))
+   (f/entry :notes c/Markdown
+            :description (str "provide any additional comments about the "
+                              "weakness"))))
+
+;; Does not contain these fields from CWE
+;; - Extended_Description
```
saintx: Our Describable field allows us to add a longer 5000 character Markdown description, so we could potentially put a lot of these unincluded fields into that if we wished.  I think the ones you decided to omit seem reasonable.

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,202 @@
...
+;; - Related_Weaknesses (Relationship)
+;; - Weakness_Ordinalities
+;; - Demonstrative_Examples
+;; - Observed_Examples
+;; - Taxonomy_Mappings
+;; - Related_Attack_Patterns
+;; - References (external_references)
```
saintx: We need to support `:external_references`, as we do for Attack Patterns (https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/attack_pattern.cljc#L32)

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,202 @@
...
+;; - Content_History
+;; - Name
+;; - Abstraction
```
saintx: We should support this as an optional field `:abstraction_level`, with the values enumeration of `["Class", "Base", "Variant",  "Compound"]` as defined https://cwe.mitre.org/documents/schema/#AbstractionEnumeration , similar to how we support it in Attack Patterns (https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/attack_pattern.cljc#L49)

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,202 @@
...
+            :description (str "identify system resources that can be affected "
+                              "by an exploit of this weakness"))
+   (f/entry :notes c/Markdown
+            :description (str "provide any additional comments about the "
+                              "weakness"))))
+
+;; Does not contain these fields from CWE
+;; - Extended_Description
+;; - Related_Weaknesses (Relationship)
+;; - Weakness_Ordinalities
```
saintx: > The WeaknessOrdinalitiesType complex type indicates potential ordering relationships with other weaknesses. A primary relationship means the weakness exists independent of other weaknesses, while a resultant relationship is when a weakness exists only in the presence of some other weaknesses. The required Ordinality element identifies whether the weakness has a primary or resultant relationship. The optional Description contains the context in which the primary or resultant relationship exists. It is important to note that it is possible for the same entry to be primary in some instances and resultant in others.

We can actually represent these weakness ordinalities in the relationship type, so we don't need them in here.  Good call not including this.

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,202 @@
...
+  c/base-entity-entries
+  c/describable-entity-entries
+  c/sourcable-object-entries
+  (f/required-entries
+   (f/entry :type WeaknessTypeIdentifier
+            :description (str "The fixed value " type-identifier))
+   (f/entry :description c/Markdown
+            :description (str "should be short and limited to the key points "
+                              "that define this weakness"))
+   (f/entry :structure v/WeaknessStructure
```
saintx: I think this should be optional.

src/ctim/schemas/weakness.cljc
msprunck: `external_references` are already included by `base-entity-entries`

saintx:
Looks good

src/ctim/schemas/weakness.cljc
```diff
@@ -0,0 +1,218 @@
...
+   (f/entry :architectures [Architecture]
+            :description "Applicable architectures")
+   (f/entry :paradigms [Paradigm]
+            :description "Applicable paradigms")
+   (f/entry :technologies [Technology]
+            :description "Applicable technologies")
+   (f/entry :background_details c/Markdown
+            :description (str "information that is relevant but not related to "
+                              "the nature of the weakness itself"))
+   (f/entry :alternate_terms [c/ShortString]
```
saintx: Alternate terms can have optional descriptions:

```
{:alternate-term
   {:term "Tapjacking",
    :description
    "\"Tapjacking\" is similar to clickjacking, except it is used for mobile applications in which the user \"taps\" the application instead of performing a mouse click."}}
```

PR #207, opened 2018-06-19
Added more structure to the Weakness Note type

gbuisson:
closes https://github.com/threatgrid/iroh/issues/1994

- [x] schema change
- [x] examples updated
- [x] doc updated

### DIFF
--- a/doc/json/bundle.json
+++ b/doc/json/bundle.json
@@ -516,7 +516,10 @@
     "timestamp" : "2016-01-01T01:01:01.000Z",
     "revision" : 10,
     "abstraction_level" : "string",
-    "alternate_terms" : [ "string" ],
+    "alternate_terms" : [ {
+      "term" : "string",
+      "description" : "string"
+    } ],
     "short_description" : "string",
     "languages" : [ {
       "prevalence" : "string",
@@ -547,7 +550,10 @@
     } ],
     "likelihood" : "string",
     "source_uri" : "string",
-    "notes" : "string",
+    "notes" : [ {
+      "type" : "string",
+      "note" : [ "string" ]
+    } ],
     "technologies" : [ {
       "prevalence" : "string",
       "name" : "string"
--- a/doc/json/bundle.json
+++ b/doc/json/bundle.json
@@ -552,7 +552,7 @@
     "source_uri" : "string",
     "notes" : [ {
       "type" : "string",
-      "note" : [ "string" ]
+      "note" : "string"
     } ],
     "technologies" : [ {
       "prevalence" : "string",
--- a/doc/json/casebook.json
+++ b/doc/json/casebook.json
@@ -566,7 +566,7 @@
       "source_uri" : "string",
       "notes" : [ {
         "type" : "string",
-        "note" : [ "string" ]
+        "note" : "string"
       } ],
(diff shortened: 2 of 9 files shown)


gbuisson:
@saintx I set the notes/note field to be a vector of markdown but at this state I'm unsure, we could also settle for a single markdown string or a hiccup form, what do you prefer ?

saintx:
I have been reducing them into a single markdown string.

src/ctim/schemas/weakness.cljc
```diff
@@ -141,6 +141,11 @@
     (f/entry :effectiveness_notes c/MedString)))
   :reference "[PotentialMitigationsType](https://cwe.mitre.org/documents/schema/#PotentialMitigationsType)")
 
+(def-map-type Note
+  (f/required-entries
+   (f/entry :type v/NoteType)
+   (f/entry :note [c/Markdown])))
```
msprunck: Why an array of `c/Markdown`? You can create paragraph directly with the Makdown format.

saintx: Yeah it definitely should not be an array of `c/Markdown`.  There's not 1 `:type` and n `:note` fields.  There are n `:type` and m `:note` fields.

gbuisson: because we can! No seriously I have no preference myself, just that the array is more premissive in the future vs a single value, I take the votes!

saintx:
Thank you for this.  I really appreciate it.

doc/json/bundle.json
```diff
@@ -547,7 +550,10 @@
     } ],
     "likelihood" : "string",
     "source_uri" : "string",
-    "notes" : "string",
+    "notes" : [ {
+      "type" : "string",
+      "note" : "string"
```
saintx: I'd make this `[string]`

doc/json/casebook.json
```diff
@@ -561,7 +564,10 @@
       } ],
       "likelihood" : "string",
       "source_uri" : "string",
-      "notes" : "string",
+      "notes" : [ {
+        "type" : "string",
+        "note" : "string"
```
saintx: I'd make this a `[string]`

gbuisson:
Markdown is internally just a plain Str, it's up to you to format it as a plain string or a Markdown String, I think it's preferable to call it markdown so we allow that formatting.

doc/json/casebook.json
```diff
@@ -530,7 +530,10 @@
       "timestamp" : "2016-01-01T01:01:01.000Z",
       "revision" : 10,
       "abstraction_level" : "string",
-      "alternate_terms" : [ "string" ],
+      "alternate_terms" : [ {
+        "term" : "string",
```
saintx: This is acceptable.  We never need more than one "term" per "alternate_terms" value.

doc/json/casebook.json
```diff
@@ -530,7 +530,10 @@
...
+        "description" : "string"
```
saintx: The `"description"` must be optional.  Many `"alternate_terms"` entries don't have any description text.

saintx:
Let's get this merged.

PR #209, opened 2018-06-20
Vulnerability Entity

### LINKED ISSUE #195: Add Vulnerability entity
Add a Vulnerability entity, based on the STIX 2.0 SDO of the same name.

- [ ] Extends `common/base-entity-entries`
- [ ] Vulnerability should be `sourceable` and `describable`.
- [ ] Add an example for the entity, if none exists.

## NVD Updates, June 21st
I compared this schema with the CVE XSD from the Mitre.org, and also crawled through the CVE dataset, and I think this schema would provide us with all of that.

However, one of the other resources that I am very interested in using is the National Vulnerability Database (NVD), an example of which can be found here:  https://nvd.nist.gov/vuln/detail/CVE-2017-0144

One of the most compelling differences between the NVD and the CVE data is that, in addition to the base CVE data, NVD also provides interesting information on the attack vector, as well as the Common Vulnerability Scoring System (CVSS) scores for each CVE.  I think this contextual information is actually very important and would be worth including.  For example, in the case of CVE-2017-0144 which I mentioned above, the attack vector string `"AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H"` can tell us that this vulnerability has a network attack vector, high attack complexity, no user interaction, and no special privileges required, in addition to some other things.  It also conveys the impact to confidentiality, integrity, and availability for a successful attack.  Given this information, and the availability of NVD data sets, I'm more inclined to pull this information directly from the NVD, rather than mitre.

I think we could augment our current model with a `:severity` score, an `:impact` assessment on `:confidentiality`, `:availability`, and `:integrity`, as well as `:attack_vector`, `:attack_complexity`, `:privileges_required`, `:user_interaction`, and`:scope`.  Also, I think a `:published_date` and `:last_modified` date would be useful.

## References
- [CVSS v3.0 Specification](https://www.first.org/cvss/cvss-v30-specification-v1.8.pdf)

gbuisson:
closes #195 

- [x] Definition
- [x] Examples
- [x] Generators
- [x] Docs


src/ctim/schemas/vulnerability.cljc
```diff
@@ -0,0 +1,48 @@
...
+  c/sourcable-object-entries
+  (f/required-entries
+   (f/entry :type VulnerabilityTypeIdentifier
+            :description (str "The fixed value " type-identifier))
+   (f/entry :description c/Markdown
+            :description
+            (str "A description that provides more details and context about the Vulnerability,"
+                 "potentially including its purpose and its key characteristics.")))
+  (f/optional-entries
+   (f/entry :name c/ShortString
```
msprunck: In [Stix 2.0](http://docs.oasis-open.org/cti/stix/v2.0/cs01/part2-stix-objects/stix-v2.0-cs01-part2-stix-objects.html#_Toc496714335) the name is required and the description optional. Why is it different?

gbuisson: Out of uniformicity with our other objects, I don't think we should stick closely to stix and make the description required as it is inherited from a describable object.

I'll make the name required though.

gbuisson:
@saintx 
I updated the schema including  all `CVSSV3` metrics along with the vector string and its spec validation, `published_date`, `last_modified` and severity, tell me if it is accurate.

src/ctim/schemas/vulnerability.cljc
```diff
@@ -0,0 +1,188 @@
...
+             :description (str "Reflects the context by which "
+                               "vulnerability exploitation is possible"))
+    (f/entry :modified_attack_vector v/VulnerabilityModifiedAttackVector
+             :description "modified attack vector")
+
+    (f/entry :attack_complexity v/VulnerabilityAttackComplexity
+             :description (str "describes the conditions beyond the attacker's "
+                               "control that must exist in order to exploit the "
+                               "vulnerability"))
+    (f/entry :modified_attack_complexity v/VulnerabilityModifiedAttackComplexity
```
saintx: > ### Modified Base Metrics
>
> These metrics enable the analyst to adjust the Base metrics according to modifications that exist 
> within the analyst’s environment. That is, if an environment has made general changes for the 
> affected software that differs in a way which would affect its Exploitability, Scope, or Impact, 
> then the environment can reflect this via an appropriately-modified, Environmental score.

I am not sure how we're going to use these modified fields, at least initially.  It could be useful in private CTIA if we built tooling to let users clone a vulnerability and then augment the cvssv3 score for their environment.  Interesting addition.

PR #213, opened 2018-07-30
Weakness CWE changes

### LINKED ISSUE #212: Weakness model updates based on CWE data
There are several updates to the weakness model that we should make, based on the contents of the CWE data.

### 1. Make `:modes_of_introduction :note` optional

I'm getting the error message: `"modes_of_introduction":[{"note":"missing-required-key"},{"note":"missing-required-key"}]` when trying to import a Weakness in local CTIA (master branch).

However, `note` is not a required key in the Weakness schema's `ModeOfIntroduction` map type: https://github.com/threatgrid/ctim/blob/18f5876bbd60b6496dd6341c6a1e6b836d3826cb/src/ctim/schemas/weakness.cljc#L21-L28

### 2. `:modes_of_introduction :note` should be Markdown

Many of the `<Modes_Of_Introduction><Note>` contents are arbitrarily nested XHTML, and need to be converted to Markdown strings.

https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/weakness.cljc#L26

### 3. `:detection_method :description` should be Markdown

Many of the `<Detection_Method><Description>` contents are XHTML, and need to be converted to Markdown strings.

https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/weakness.cljc#L110

gbuisson:
closes #212 

- changed `Note`and `Description`field types to markdown
- `ModeOfIntroduction/Note` is now optional 
- align the reader dependency with our other projects (fixes a reader error)

### DIFF
--- a/src/ctim/schemas/weakness.cljc
+++ b/src/ctim/schemas/weakness.cljc
@@ -19,13 +19,15 @@
   "[WeaknessType](https://cwe.mitre.org/documents/schema/#WeaknessType)")
 
 (def-map-type ModeOfIntroduction
-  [(f/entry :phase v/SoftwarePhase
-            :description (str "identifies the point in the software life cycle "
-                              "at which the weakness may be introduced")
-            :required? true)
-   (f/entry :note c/MedString
-            :description (str "provides a typical scenario related to "
-                              "introduction during the given phase"))]
+  (concat
+   (f/required-entries
+    (f/entry :phase v/SoftwarePhase
+             :description (str "identifies the point in the software life cycle "
+                               "at which the weakness may be introduced")))
+   (f/optional-entries
+    (f/entry :note c/Markdown
+             :description (str "provides a typical scenario related to "
+                               "introduction during the given phase"))))
   :reference "[ModesOfIntroductionType](https://cwe.mitre.org/documents/schema/#ModesOfIntroductionType)")
 
 (def applicable-platform-entries
@@ -81,7 +83,7 @@
     (f/entry :term c/ShortString
              :description "the actual alternate term"))
    (f/optional-entries
-    (f/entry :description c/MedString
+    (f/entry :description c/Markdown
              :description (str "provides context for the alternate term "
                                "by which this weakness may be known.")))))
 
@@ -97,7 +99,7 @@
     (f/entry :likelihood v/HighMedLow
              :description (str "how likely the specific consequence is expected "
                                "to be seen relative to the other consequences"))
-    (f/entry :note c/MedString
+    (f/entry :note c/Markdown
              :description "additional commentary about a consequence")))
   :reference "[CommonConsequencesType](https://cwe.mitre.org/documents/schema/#CommonConsequencesType)")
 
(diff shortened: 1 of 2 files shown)


src/ctim/schemas/weakness.cljc
```diff
@@ -97,7 +99,7 @@
     (f/entry :likelihood v/HighMedLow
              :description (str "how likely the specific consequence is expected "
                                "to be seen relative to the other consequences"))
-    (f/entry :note c/MedString
+    (f/entry :note c/Markdown
```
saintx:
```
(->> "/var/cwe/1000.xml"
       input-stream
       xml-input-stream->weakness-records
       (map :common-consequences)
       (mapcat identity)
       (map :consequence)
       (map :note)
       (remove nil?)
       (map type)
       set) => #{java.lang.String}
```
This change is not needed, there are no cases in the production data where it uses anything other than a simple string.

src/ctim/schemas/weakness.cljc
```diff
@@ -107,14 +109,14 @@
...
-    (f/entry :description c/MedString
+    (f/entry :description c/Markdown
              :description (str "provide some context of how this method can "
                                "be applied to a specific weakness")))
    (f/optional-entries
     (f/entry :effectiveness v/DetectionEffectiveness
              :description (str "how effective the detection method may be in "
                                "detecting the associated weakness"))
-    (f/entry :effectiveness_notes c/MedString
+    (f/entry :effectiveness_notes c/Markdown
```
saintx:
```
(->> "/var/cwe/1000.xml"
       input-stream
       xml-input-stream->weakness-records
       (map :detection-methods)
       (mapcat identity)
       (map :detection-method)
       (map :effectiveness-notes)
       (remove nil?)
       (map type)
       set) => #{java.lang.String}
```
This change to `:effectiveness_notes` is not needed, there are no cases in the production data where it uses anything other than a simple string.

src/ctim/schemas/weakness.cljc
```diff
@@ -138,7 +140,7 @@
     (f/entry :effectiveness v/Effectiveness
              :description (str "summarizes how effective the mitigation may be "
                                "in preventing the weakness"))
-    (f/entry :effectiveness_notes c/MedString)))
+    (f/entry :effectiveness_notes c/Markdown)))
```
saintx:
```
(->> "/var/cwe/1000.xml"
       input-stream
       xml-input-stream->weakness-records
       (map :potential-mitigations)
       (mapcat identity)
       (map :mitigation)
       (map :effectiveness-notes)
       (remove nil?)
       (map type)
       set) => #{java.lang.String}
```
This change to `:effectiveness_notes` is not needed here, there are no cases in the production data where it uses anything other than a simple string.  The change to the `:description` IS needed, though!

src/ctim/schemas/weakness.cljc
```diff
@@ -81,7 +83,7 @@
     (f/entry :term c/ShortString
              :description "the actual alternate term"))
    (f/optional-entries
-    (f/entry :description c/MedString
+    (f/entry :description c/Markdown
```
saintx: This change was definitely needed, thank you.

src/ctim/schemas/weakness.cljc
```diff
@@ -19,13 +19,15 @@
...
-   (f/entry :note c/MedString
-            :description (str "provides a typical scenario related to "
-                              "introduction during the given phase"))]
+  (concat
+   (f/required-entries
+    (f/entry :phase v/SoftwarePhase
+             :description (str "identifies the point in the software life cycle "
+                               "at which the weakness may be introduced")))
+   (f/optional-entries
+    (f/entry :note c/Markdown
```
saintx: Perfect, this needed to be both optional, and Markdown.

gbuisson:
I'm surprised you're blocking on that,
as I said on the PR opening comment, I don't think we want to operate on a per field basis for fields of the same type, I know we generate good documentation with flanders, however I think people should also be able to rely on a field name to guess the type, if one encounter `note` and it's markdown then one should expect all other similar fields with a similar name to be of the same type, moreover markdown can contain a string just fine, under the hood just the field max length changes: https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/common.cljc#L143

PR #219, opened 2018-10-16
Add odns_identity_label observable type

msprunck:
Connected to threatgrid/iroh#2162

Add `odns_identity_label` observable type that matches the `originLabel` field in the Umbrella Reporting API reponses.

### DIFF
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -258,7 +258,8 @@
     "mac_address"
     "file_name"
     "file_path"
-    "odns_identity"})
+    "odns_identity"
+    "odns_identity_label"})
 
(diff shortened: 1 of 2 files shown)


src/ctim/schemas/vocabularies.cljc
```diff
@@ -258,7 +258,8 @@
     "mac_address"
     "file_name"
     "file_path"
-    "odns_identity"})
+    "odns_identity"
+    "odns_identity_label"})
 
 (def-enum-type ObservableTypeIdentifier
   observable-type-identifier
```
yogsototh: I think that if we're adding many new specific observable type for different product/context we should make the "ObservableTypeIdentifier" open to any string.

hshahman:
Verified in https://github.com/threatgrid/iroh/pull/2196

PR #224, opened 2018-12-04
Remove all Stored* schemas from CTIM

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2176

They don't belong to this repo and should be derived generically in CTIA


src/ctim/schemas/common.cljc
```diff
@@ -23,7 +23,18 @@
...
+
+(def SchemaVersion
+  (f/str
+   :description (str "A semantic version matching the CTIM version against which "
+                     "this object should be valid.")
+   :default ctim-schema-version
+   :gen (cs/gen #{ctim-schema-version})
+   :spec (cs/and string? ::ctim-schema-version)))
+
+(def CTIMSchemaVersion SchemaVersion)
```
msprunck: Why did you change the schema of the clim-schema-version? For me the value for each CTIM version if a static value.

PR #228, opened 2019-02-11
Add Registry Keyval Pair, Socket Address

saintx:
This update unites the "registry_key" and "registry_val" into a single
canonical "registry_keyval" observable type, and also adds support for
socket addresses, which are the combination of an IP address and a port number.

Added documentation on the `registry_keyval` observable type, to document its construction.

### DIFF
--- a/doc/observables/registry_keyval.md
+++ b/doc/observables/registry_keyval.md
@@ -1,0 +1,103 @@
+# The `registry_keyval` Observable Type
+
+## Introduction
+
+This guide will document how to combine Registry Key and Registry Value Entries into a single canonical string format which can be used as an observable value in CTIM.
+
+### Registry Keys
+
+The Windows Registry is a file database that allows multiple **Value Entries** to be stored under a **Key Path**.  The registry supports seven root keys, some of which have strictly controlled sets of officially designated subkeys, into which additional subkeys can be nested.
+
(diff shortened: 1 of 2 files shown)


doc/observables/registry_keyval.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+
+## Canonical String Format
+
+We can encode the registry key and value entry using the following canonical form:
+
+```
+registry_kv: "HKEY_CURRENT_USER\SOFTWARE\MICROSOFT\WINDOWS\CURRENTVERSION\RUN\\\"api-PQEC\"=REG_DWORD:0x800000"
+```
```
msprunck: Maybe it's time to change the structure of the `Observable` value object by allowing more than a `type` and a `value`. [STIX 2](http://docs.oasis-open.org/cti/stix/v2.0/cs01/part4-cyber-observable-objects/stix-v2.0-cs01-part4-cyber-observable-objects.html#_Toc496716290) encodes Windows Registry keys in a more structured way. I prefer their approach, it's more readable by a human or a machine.

saintx: I had considered using combinations of Observables, but not adding optional fields  to the Observable value object.  I'm in favor of holding off on this until @craigbro has a chance to weigh in.

PR #229, opened 2019-02-13
added observables ip helpers

ereteog:
related to https://github.com/threatgrid/iroh/issues/2468

Introduces helpers for ip observables.

### DIFF
--- a/src/ctim/domain/observables/ip.clj
+++ b/src/ctim/domain/observables/ip.clj
@@ -4,6 +4,11 @@
             [schema.core :as s]
             [clojure.network.ip :refer [make-network make-ip-address]]))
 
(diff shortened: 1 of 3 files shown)


src/ctim/domain/observables/ip.clj
```diff
@@ -0,0 +1,90 @@
...
+  [ip-str :- s/Str
+   mask :- s/Str]
+  (-> (make-network mask)
+      (contains? ip-str)))
+
+(s/defn match-some-masks? :- s/Bool
+  "Does this ip match one of given masks?"
+  [ip-str :- s/Str
+   masks :- [s/Str]]
+  (true? (some (partial match-mask? ip-str) masks)))
```
gbuisson: I generally use `boolean` but `true?` just works here I think

ereteog: I have used `true?` because [`some`](https://clojuredocs.org/clojure.core/some) "returns the first logical `true` value of (pred x) for any x in coll, else `nil`".

PR #238, opened 2019-04-12
Adding  Sensor Coordinates for Sighting

jyoverma:
1. Added Sensor Coordinates in Sighting to hold details about the sensor that made the sighting such as ip, mac address etc. 
2. Added ngfw_id, ngfw_name to ObservableType vocabulary
3. Renamed SightingTarget to TargetCoordinates

No QA required

### DIFF
--- a/src/ctim/schemas/sighting.cljc
+++ b/src/ctim/schemas/sighting.cljc
@@ -17,15 +17,23 @@
       :refer-macros
       [def-entity-type def-eq def-map-type]])]))
 
-(def-map-type SightingTarget
+(def-map-type TargetCoordinates
   (concat
    (f/required-entries
     (f/entry :type v/Sensor)
     (f/entry :observables [c/Observable])
     (f/entry :observed_time c/ObservedTime))
    (f/optional-entries
-    (f/entry :os f/any-str)
-    (f/entry :properties_data_tables rel/DataTableReference)))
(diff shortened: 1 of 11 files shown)


src/ctim/schemas/sighting.cljc
```diff
@@ -17,15 +17,23 @@
...
+  :description "Describes a target device where a sighting came from.")
+
+(def-map-type SensorCoordinates
+  (concat
+   (f/required-entries
+    (f/entry :type v/Sensor)
+    (f/entry :observables [c/Observable]))
+   (f/optional-entries
+    (f/entry :os f/any-str)))
   :description "Describes a target device where a sighting came from.")
```
craigbro: This description should be updated.  We should also make it more appropriate for coordinates.  "Contains the identifying observables or a target|sensor..."

src/ctim/schemas/vocabularies.cljc
```diff
@@ -260,6 +260,8 @@
     "file_path"
     "odns_identity"
     "odns_identity_label"
+    "ngfw_id"
```
craigbro: Let's confirm with the SSE team that this is the appropriate name

gbuisson:
Also please remember that when you update a model, you need to also update the examples and tests if needed.

PR #246, opened 2019-06-25
New ObservablePattern specification for Indicator model

jyoverma:
Added Specification type called ObservablePattern for Indicator data model


src/ctim/schemas/indicator.cljc
```diff
@@ -41,6 +41,21 @@
    (f/entry :snort_sig f/any-str))
   :description "An indicator which runs in snort...")
 
+(def-map-type Pattern
```
gbuisson: here you fix indentation

jyoverma:
This change is not needed at this time since the integration with CTA got revised

PR #247, opened 2019-07-01
Harden an ID conversion helper fn to help avoid a NPE

gbuisson:
Title is self explaining

connected to https://github.com/threatgrid/iroh/issues/2810

### DIFF
--- a/src/ctim/domain/id.cljc
+++ b/src/ctim/domain/id.cljc
@@ -135,7 +135,7 @@
   [short-id :- (s/maybe s/Str)
    url-params-fn :- (s/maybe (s/=> {s/Keyword s/Any}))]
   (when short-id
-    (when-let [type (nth (re-matches short-id-re short-id) 2)]
+    (when-let [type (nth (re-matches short-id-re short-id) 2 nil)]
       (long-id
        (short-id->id type
                      short-id
--- a/src/ctim/domain/id.cljc
+++ b/src/ctim/domain/id.cljc
@@ -131,12 +131,15 @@
     (nth (re-matches long-id-re s) 8)
     s))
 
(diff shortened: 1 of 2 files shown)


src/ctim/domain/id.cljc
```diff
@@ -131,12 +131,15 @@
...
-(defn short-id->long-id [short-id url-params-fn]
-  (when-let [type (nth (re-matches short-id-re short-id) 2)]
-    (long-id
-     (short-id->id type
-                   short-id
-                   (url-params-fn)))))
+(s/defn short-id->long-id :- (s/maybe s/Str)
+  [short-id :- (s/maybe s/Str)
+   url-params-fn :- (s/maybe (s/=> {s/Keyword s/Any}))]
+  (when short-id
```
yogsototh: Wouldn't it be better to log when `short-id` is `nil` because it might be a bug?

msprunck: IMO it should not be logged in CTIM, it's like calling `map` with a `nil` coll. It should just return `nil`

src/ctim/domain/id.cljc
```diff
@@ -131,12 +131,15 @@
...
+    (when-let [type (nth (re-matches short-id-re short-id) 2)]
```
ereteog: you could remove first `when` and use [`not-found`parameter of `nth`](https://clojuredocs.org/clojure.core/nth):
`nth throws an exception unless not-found is supplied.`
```clojure
(when-let [type (nth (re-matches short-id-re short-id) 2 nil)]
...
)
```

gbuisson:
I'm with Matt, I think logging should be done at levels above

PR #252, opened 2019-09-04
Add Orbital fields to CTIM vocabulary

jyoverma:
Adding orbital_id and registry_key to observable vocabulary

### DIFF
--- a/src/ctim/examples/indicators.cljc
+++ b/src/ctim/examples/indicators.cljc
@@ -39,14 +39,12 @@
                        {:kill_chain_name "kcn2"
                         :phase_name "persistence"}]
    :test_mechanisms ["spam" "eggs"]
-   :specification {:type "ObservablePattern"
-                   :pattern {:type "url" :value "(.*\\.)?mybaddomain.com/.*/(acttr|com|cmn2|tbd|vast\\\\.php|hks|pubjs|imp|drt3).*"}}})
-  ;  :specification {:type "Judgement"
-  ;                  :judgements ["http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"]
-  ;                  :required_judgements [{:judgement_id "http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"
-  ;                                         :source "source"
-  ;                                         :relationship "relationship"
-                                          ; :confidence "High"}]}})
+   :specification {:type "Judgement"
+                   :judgements ["http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"]
+                   :required_judgements [{:judgement_id "http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"
+                                          :source "source"
+                                          :relationship "relationship"
+                                          :confidence "High"}]}})
 
 (def indicator-minimal
   {:id "http://ex.tld/ctia/indicator/indicator-10b490f8-7c1d-4c3b-9be3-52a7fa5c5042"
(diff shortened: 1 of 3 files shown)


src/ctim/schemas/vocabularies.cljc
```diff
@@ -265,7 +266,8 @@
     "email_messageid"
     "email_subject"
     "cisco_mid"
-    "mutex"})
+    "mutex"
+    "registry_key"})
```
msprunck: The `registry_key` and `registry_value` observable types have been removed in this PR https://github.com/threatgrid/ctim/pull/230. Did you discuss about it with @saintx?

src/ctim/schemas/vocabularies.cljc
```diff
@@ -254,6 +254,7 @@
     "imei"
     "imsi"
     "amp_computer_guid"
+    "orbital_id"
```
msprunck: What is an `orbital_id`? I have a doubt about it being an observable

src/ctim/examples/indicators.cljc
```diff
@@ -39,12 +39,14 @@
...
-                                          :relationship "relationship"
-                                          :confidence "High"}]}})
+   :specification {:type "ObservablePattern"
+                   :pattern {:type "url" :value "(.*\\.)?mybaddomain.com/.*/(acttr|com|cmn2|tbd|vast\\\\.php|hks|pubjs|imp|drt3).*"}}})
+  ;  :specification {:type "Judgement"
+  ;                  :judgements ["http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"]
+  ;                  :required_judgements [{:judgement_id "http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"
+  ;                                         :source "source"
+  ;                                         :relationship "relationship"
+                                          ; :confidence "High"}]}})
```
msprunck: Could you remove the Judgement specification if it's no longer used?

src/ctim/schemas/indicator.cljc
```diff
@@ -41,6 +41,21 @@
    (f/entry :snort_sig f/any-str))
   :description "An indicator which runs in snort...")
 
+(def-map-type Pattern
+(f/required-entries
```
msprunck: bad indentation here

src/ctim/schemas/indicator.cljc
```diff
@@ -41,6 +41,21 @@
...
+ (f/entry :type f/any-str)
```
msprunck:
```suggestion
 (f/entry :type v/ObservableTypeIdentifier)
```

src/ctim/schemas/indicator.cljc
```diff
@@ -41,6 +41,21 @@
...
+ (f/entry :value f/any-str)))
```
msprunck: Could you define a type with a spec and a generator that ensures the validity of the regex?
Ex: https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/common.cljc#L31

src/ctim/schemas/vocabularies.cljc
jyoverma: Yes, we talked and since the PR that was to re-introduce the registry types is on hold because of the inability to support composite observables, we can add this back for the Orbital CTR integration

src/ctim/schemas/vocabularies.cljc
jyoverma: This is the identifier of the orbital node

src/ctim/schemas/indicator.cljc
jyoverma: Actually ObservablePattern specification is not needed at this time since there is no clear use case for it. Will remove it for now. It was added when the CTA integration was being discussed but that got dropped.

src/ctim/schemas/vocabularies.cljc
saintx: The structured registry key path, data type, and value that I describe in https://github.com/threatgrid/hydrant/wiki/Encoding-Registry-Key-Values will have to wait.  We can call it something else.  We need some kind of `registry_key` observable type for orbital, so we should add it.

PR #253, opened 2019-09-09
avoid exceptions when testing special ips

ereteog:
This PR avoid throwing an exception when calling `special-ip?` or `private-ip?` with an invalid ip. These functions, now simply return `false` when the given ip is invalid.
> related https://github.com/threatgrid/iroh/issues/2588

### DIFF
--- a/src/ctim/domain/observables/ip.clj
+++ b/src/ctim/domain/observables/ip.clj
@@ -1,6 +1,5 @@
 (ns ctim.domain.observables.ip
   (:require [clojure.string :as string]
-            [clojure.pprint :refer [pprint]]
             [schema.core :as s]
             [clojure.network.ip :refer [make-network make-ip-address]]))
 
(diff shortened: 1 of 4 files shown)


src/ctim/domain/observables/ip.clj
```diff
@@ -54,18 +53,36 @@
    "2620:4f:8000::/48"
    "fe80::/10"])
 
+(s/defn valid-ipv4? :- s/Bool
+  "Is this a valid ipv4 address?"
+  [ip-str :- s/Str]
+  (some? (re-matches ipv4-regex ip-str)))
```
msprunck: could you also handle the `nil` value to avoid NPE?

src/ctim/domain/observables/ip.clj
```diff
@@ -54,18 +53,38 @@
    "2620:4f:8000::/48"
    "fe80::/10"])
 
+(s/defn valid-ipv4? :- s/Bool
+  "Is this a valid ipv4 address?"
+  [ip-str :- s/Str]
+  (and (string? ip-str)
```
yogsototh: I think it would be nicer to log when those function are wrongly used.

```suggestion
  (and (or (string? ip-str) (log/error "valid-ipv4? called with a non string value"))
```

Issue #261, opened 2019-11-26
Add new indicator type for custom indicators


jyoverma:
Add a new indicator type called "PrivateThreatFeed" for user created indicators.

PR #263, opened 2019-11-29
Adding new  indicator type for custom feed indicators

jyoverma:
Renaming new indicator type as "Private Threat Feed"


jyoverma:
#Closes https://github.com/threatgrid/ctim/issues/261

PR #269, opened 2020-01-31
Update dependencies and test JDK 11

### LINKED ISSUE #268: Update dependencies and test JDK 11

frenchy64:
Close #268 

Passes iroh tests https://github.com/threatgrid/iroh/pull/3143
Passes ctia tests https://github.com/threatgrid/ctia/pull/871


src/ctim/generators/events.clj
```diff
@@ -1,3 +1,4 @@
+(comment
```
msprunck: Why this ns is in comment? If not used, please remove it

test/ctim/test_helpers/core.cljc
```diff
@@ -27,9 +23,11 @@
     (binding [cs/*recursion-limit* n]
       (t))))
 
+; this function used to override gen/vector with
+; a function with different behavior, which broke
+; internal generator logic in test.check 0.10.0.
 (defn fixture-fast-gen [t]
```
msprunck:
```suggestion
(defn ^{:deprecated "1.0.16"} fixture-fast-gen [t]
```

frenchy64: `fixture-fast-gen` is still used in just about every `ctim` test namespace (it just does nothing now). I think deprecating it can wait because it would mean deleting all the references and this PR is big and tangled enough.

src/ctim/generators/events.clj
frenchy64: Removed the comment, thanks.

test/ctim/test_helpers/core.cljc
frenchy64: Opened an issue to do later https://github.com/threatgrid/ctim/issues/270

src/ctim/generators/events.clj
frenchy64: Decided to remove as it doesn't even compile.

PR #282, opened 2020-03-04
New investigation fields

noprompt:
This adds new fields to the `Investigation` schema see [here](https://github.com/threatgrid/iroh-engine/issues/202) and [here](https://github.com/threatgrid/iroh-engine/issues/255) for background.

### DIFF
--- a/src/ctim/schemas/investigation.cljc
+++ b/src/ctim/schemas/investigation.cljc
@@ -16,8 +16,8 @@
   (f/required-entries
    (f/entry :type InvestigationIdentifier))
   (f/optional-entries
-   (f/entry :object-ids (f/seq-of (f/str)))
-   (f/entry :investigated-observables (f/seq-of c/Observable))
+   (f/entry :object_ids (f/seq-of (f/str)))
+   (f/entry :investigated_observables (f/seq-of c/Observable))
    (f/entry :targets (f/seq-of c/IdentitySpecification))))
 
 (def-entity-type NewInvestigation
--- a/src/ctim/schemas/investigation.cljc
+++ b/src/ctim/schemas/investigation.cljc
@@ -14,7 +14,10 @@
   c/sourced-object-entries
   c/describable-entity-entries
   (f/required-entries
-   (f/entry :type InvestigationIdentifier)))
+   (f/entry :type InvestigationIdentifier)
+   (f/entry :ids (f/seq-of (f/str)))
+   (f/entry :investigated-observables (f/seq-of c/Observable))
+   (f/entry :targets (f/seq-of c/IdentitySpecification))))
 
 (def-entity-type NewInvestigation
   "Schema for submitting new Investigations"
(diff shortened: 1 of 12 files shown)


src/ctim/schemas/investigation.cljc
```diff
@@ -14,7 +14,12 @@
   c/sourced-object-entries
   c/describable-entity-entries
   (f/required-entries
-   (f/entry :type InvestigationIdentifier)))
+   (f/entry :type InvestigationIdentifier))
+  (f/optional-entries
+   (f/entry :object_ids (f/seq-of (f/str)))
+   ;; "type:value"
+   (f/entry :investigated_observables (f/seq-of (f/str)))
```
msprunck: Why do you prefer the string format for an observable? It will require additional parsing

src/ctim/schemas/investigation.cljc
```diff
@@ -14,7 +14,12 @@
...
+   (f/entry :targets (f/seq-of c/IdentitySpecification))))
```
msprunck: Could you add a description to these new fields?

Ex:
```clojure
(f/entry :targets (f/seq-of c/IdentitySpecification)
         :description "Investigated target devices")
```

src/ctim/schemas/investigation.cljc
```diff
@@ -14,7 +14,12 @@
   c/sourced-object-entries
   c/describable-entity-entries
   (f/required-entries
-   (f/entry :type InvestigationIdentifier)))
+   (f/entry :type InvestigationIdentifier))
+  (f/optional-entries
```
msprunck: Why `:actions` is not defined in CTIM?

src/ctim/schemas/investigation.cljc
```diff
@@ -14,7 +14,12 @@
...
+   (f/entry :object_ids (f/seq-of (f/str)))
```
msprunck:
```suggestion
   (f/entry :object_ids (f/seq-of c/Reference))
```

src/ctim/schemas/investigation.cljc
noprompt: I'm not sure but I can add that.

src/ctim/schemas/investigation.cljc
noprompt: Originally, this is what I had in the initial commits. In a conversation with between @craigbro, @jbouwman, and I regarding these new fields and in the context of ES and CTIA, it was suggested the data for this field be stored as `"type:value"` rather than the full observable or a `{:type "value"}`. This is why the schema is this way. Personally, I have no preference; I'm just interested in doing the "right" thing.

src/ctim/schemas/investigation.cljc
noprompt: Yes. I'm still working on this but I'll apply this suggestion.

gbuisson:
After reviewing IMHO, I believe that this entity definition should be taken out of CTIM and moved to CTIA,
it was OK at first to let it be here because it was very abstract, but now its definition exposes too much implementation details and is not relevant to the Open Source CTIM repository.

saintx:
I think that closing this was premature, and making Joel figure out the nuances of how to perform schema validation with entities that are outside of CTIM is scope creep.  If the Investigation entity needs to be removed from CTIM, that should be its own separate issue, and not impede this work.

PR #289, opened 2020-07-15
Basic Asset model

agzam:
Across incident response, network policy and access control there is a
common set of assets which are being protected.

This is a basic scaffolding to build (and expand later) Asset entity model.


jyoverma:
The asset types should be as described in assets.org
device
person
network
data
application

PR #292, opened 2020-07-17
Asset: bundle changes

agzam:
An AssetMapping is a record that a specific Observable maps to an Asset for a specific time period.


src/ctim/examples/assets.cljc
```diff
@@ -35,3 +35,53 @@
...
+(def asset-mapping-maximal
+  {:id                  "http://ex.tld/ctia/asset-mapping/asset-mapping-668d86a2-02c1-499f-b953-5383be954aa6"
+   :type                "asset_mapping"
+   :schema_version      c/ctim-schema-version
+   :revision            1
+   :external_ids        ["http://ex.tld/ctia/asset-mapping/asset-mapping-636ef2cc-1cb0-47ee-afd4-ecc1fe4be451"]
+   :description         "asset mapping"
+   :short_description   "mapping-shmaping"
+   :title               "mapping-to CBROZEFS-DH-123"
+   :source              "cisco:unified_connect"
```
rplevy, line 52: I've seen this source mentioned in documentation, and seeing it here again reminds me that I was wondering what vocabulary does this originate from?  Or do I have the wrong idea that it should be found in a vocabulary of some sort?

agzam: Good question. `source` makes sense for things like sightings, where you required to link the source where the alert data originated. I am really not sure if Asset and AssetMapping do need to have a `source`.

rplevy: I know what source is for, my question is about the value of source.

rplevy: Can the value of source be anything or does it have to be defined in a vocabulary?

jyoverma: source doesn't come from a vocabulary. Its the producer of the content

agzam:
@rplevy @jyoverma I have separated AssetMappings and Asset. The reason I put them in the same namespace - I needed to figure out common parts of schemas, right now they are minimal, I hope it stays that way as we add more Asset related things.

PR #298, opened 2020-08-17
Target Record

agzam:
A TargetRecord is a Sighting that has no threat or observables associated with it, it’s a way of saying they saw a set of observables together as a Target. This is useful when attempting to determine what assets were targeted by a threat when we only have weak transient observables, like IPs or dynamically assigned hostnames.

<a name="qa">[§](#qa)</a> QA
============================

Target Record context should be visible in Swagger UI.

Basic CRUD operations should work.


ereteog (approved):
LGTM!

PR #300, opened 2020-08-20
asset_mapping: align enums with existing vocabulary

ereteog:
This PR proposes a slight modification of the vocabulary for `confidence`, `specificity` and `stability`:
- there is an existing vocabulary for `confidence` which, in addition of `High`/ `Medium` / `Low`, also contains `Unique`, `Info` and `Unknow`. This PR proposes to reuse it and to preserve the existing case: https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/vocabularies.cljc#L73
- existing vocabularies like `confidence` use a capitalized first letter, this PR proposes to follow this "convention".

### DIFF
--- a/src/ctim/examples/asset_mappings.cljc
+++ b/src/ctim/examples/asset_mappings.cljc
@@ -19,14 +19,13 @@
    :language            "language"
    :tlp                 "green"
 
-   :confidence  "high"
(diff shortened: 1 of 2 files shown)


msprunck (approved):
LGTM!

PR #310, opened 2020-09-21
Issue-442: Add ms_machine_id and s1_agent_id observable types

halcyon:
Necessary to add support for third party host ID observables - https://github.com/threatgrid/hydrant/issues/442

This `adds ms_machine_id` and `s1_agent_id` observable types.

### DIFF
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -255,6 +255,8 @@
     "imei"
     "imsi"
     "amp_computer_guid"
+    "ms_machine_id" ;; Microsoft Windows Defender Machine ID
+    "s1_agent_id" ;; Sentinel One Agent ID
     "orbital_node_id"
     "hostname"
     "mac_address"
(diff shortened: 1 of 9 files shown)


src/ctim/schemas/vocabularies.cljc
```diff
@@ -255,6 +255,8 @@
     "imei"
     "imsi"
     "amp_computer_guid"
+    "ms_machine_id"
+    "s1_agent_id"
```
msprunck, line 259: Could you add a comment with a description or a reference link? What is a `ms_machine_id`? `s1_agent_id`?

msprunck (approved):
LGTM!

src/ctim/schemas/vocabularies.cljc
halcyon: Thank you - I don't know much about these beyond what I've managed to determine by examining the examples in the issue. 

Here are my notes:
> Microsoft Windows Defender machine IDs appear to be 40 character hexadecimal strings
> Sentinel One Agent IDs appear to be 18 character decimal strings

As I'm far from an authoritative source, would it make sense to add these as a comment to `vocabularies.cljc`?

halcyon: Will do!

PR #311, opened 2020-09-21
Issue 434: Adding name, short_description to judgement and sightings.

rubygeek:


src/ctim/schemas/judgement.cljc
```diff
@@ -47,7 +47,9 @@
    (f/entry :priority Priority)
    (f/entry :confidence v/HighMedLow)
    (f/entry :severity v/HighMedLow)
-   (f/entry :valid_time c/ValidTime))
+   (f/entry :valid_time c/ValidTime)
+   (f/entry :title c/ShortString)
+   (f/entry :short_description c/MedString)))
```
msprunck, line 52: Not sure we want a `title` and the `description` because the `reason` field is here to explain the reason for issuing the judgement, or to provide additional context.

gbuisson:
Hi, IMHO we have 2 dependencies so we can merge this PR:

- Prepare the CTIA migration so we can ship a CTIM update in case of an emergency.
- Coordinate with Maribell and the Remote Module teams to schedule any required upgrades to comply with the updated schemas.

src/ctim/schemas/judgement.cljc
saintx: I agree. This is a request from Craig and I'm not sure it's actually needed at all. The UI wants to put a title and description in for judgements, but I'm not really sure what those would be.

PR #320, opened 2021-01-21
make actor, attack-pattern, campaign, malware and tool  described entitties

ereteog:
> related #https://github.com/threatgrid/iroh/issues/4058

Adds `described-entity-entries`, the required fields couterpart of `describable-entity-entries`.
Make `described-entity-entries` to `malware`, `tool` and `attack-pattern` entities, by replacing `name` by `title`.
Replace `describable-entity-entries` by `described-entity-entries` in `campaign` and `actor`, which makes `title`, `description` and `short_description` required.

A PR is coming in CTIA with corresponding migration (concerned indices are small).

### DIFF
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -20,7 +20,7 @@
             [flanders.predicates :as fp]
             [clojure.string :as str]))
 
-(def ctim-schema-version "1.0.24")
+(def ctim-schema-version "1.1.0")
 
 (def-eq CTIMSchemaVersion ctim-schema-version)
 
@@ -221,6 +221,15 @@
              :description "A valid entity type identifer")
     (f/entry :schema_version CTIMSchemaVersion
              :description "CTIM schema version for this entity"))))
+
+(def described-entity-entries
+  "Fields for described entities"
+  [(f/entry :title ShortString
+           :description "A short title for this object, used as primary display and reference value")
+  (f/entry :description Markdown
+           :description "A description of object, which may be detailed.")
+  (f/entry :short_description MedString
+           :description "A single line, short summary of the object.")])
 
 (def describable-entity-entries
   "These fields for describable entities"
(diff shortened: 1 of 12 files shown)


src/ctim/schemas/common.cljc
```diff
@@ -20,7 +20,7 @@
             [flanders.predicates :as fp]
             [clojure.string :as str]))
 
-(def ctim-schema-version "1.0.24")
+(def ctim-schema-version "1.1.0")
```
ereteog, line 23: these changes introduce incompatibilities without being significant breaking changes.

vemv: Maybe at some point we could look into a strict application of semver. In the end there are infinite numbers so we won't run out of them :) i.e. it's simpler to be strict/pessimistic about what constitutes a breaking change

For example, https://github.com/threatgrid/ctia/pull/1065 requires a somewhat delicate migration plan. That would be even clearer if its ctim change moved from 1.* to 2.*

(in that specific PR it doesn't matter a lot since it already says `do not merge`, but maybe in a future, newer PRs could make a mistake)

ereteog: @gbuisson wdyt?

gbuisson: for 2.0.0 I would expect major changes, including for instance support for Stix 3.0, a mali output or something like that, a 1.5.0 sounds more reasonable... but I agree it's completely psychological.

msprunck: In the CTIM library, the `ctim-schema-version` should only change for schema changes, it means that even an increment in the patch version means that a breaking change like a new optional field has been added. We were used to increment only the patch or the minor version depending on the number of changes, if new entities were added. `1.1.0` looks good regarding this change. I'm open to follow another rules for the schema versioning of CTIM

vemv (approved):

hisans:
QA Verified in 1.66 Release

PR #322, opened 2021-02-01
Vulnerability CVSSv3 Update

michaels-den:
CVSS Version 3 Strings are now formatted as follows-
'CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H'
resulting in failures in the validation of this string during
vulnerability posting via hydrant importers.

Current validation expects exclusivly 'CVSS:3.0...'. This work
allows for either `3.0` or `3.1` for the version and adds an
example.

Hydrant Issue 536
IROH Issue 4729

### DIFF
--- a/src/ctim/schemas/vulnerability.cljc
+++ b/src/ctim/schemas/vulnerability.cljc
@@ -37,7 +37,7 @@
    #?(:clj gen/score)))
 
 (def cvss-v3-vector-string-exp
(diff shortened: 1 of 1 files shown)


src/ctim/examples/vulnerabilities.cljc
```diff
@@ -123,3 +123,6 @@
 
 (def new-vulnerability-minimal
   {:description "Improper Neutralization of Directives in Dynamically Evaluated Code ('Eval Injection')"})
+
+(def vulnerability-maximal-v3p1
+  (assoc-in vulnerability-maximal [:impact :cvss_v3 :vector_string] "CVSS:3.1/AV:L/AC:L/PR:H/UI:N/S:U/C:H/I:H/A:H"))
```
ereteog, line 128: We generally just provide a minimal and maximal example, but do not add other examples for different values.

ereteog (commented):
I would remove the additional example.

gbuisson (approved):
LGTM

src/ctim/examples/vulnerabilities.cljc
michaels-den: Can do - thanks!

PR #328, opened 2021-02-08
Document asset entities

agzam:
Closes: https://github.com/threatgrid/iroh/issues/4686

Document Asset, AssetProperties, AssetMapping, TargetRecord entities

## No QA required


ereteog (commented):
You did not understand my explanations :-).
Let me try to rephrase it. When you add an `entry` to a flanders schema, you use the `entry` fn that takes the name of the field and its type as parameter and you can pass 2 other parameters, `:description` and a `string`, to document this entry.
Here is a concrete example from the [incident schema](https://github.com/threatgrid/ctim/blob/f9e2eff38e51ca0c2b8c4099457a1ecf07c8dba3/src/ctim/schemas/incident.cljc#L34):
```clojure
(f/entry :confidence v/HighMedLow
         :description (str "level of confidence held in the characterization "
                           "of this Incident"))
```
This task consists in adding this `description` parameters to document the fields of the asset entities. So for instance to take a very similar entry in `asset-mapping`:
``` clojure
(f/entry :confidence v/HighMedLow)
```
you can change it to:
``` clojure
(f/entry :confidence v/HighMedLow
             :description (str "level of confidence held in the characterization "
                               "of this AssetMapping"))
```
And then you continue with all other field. Then you can regenerate the doc with `lein doc` like you did to have well docuemented fields for asset entities.

project.clj
```diff
@@ -6,7 +6,7 @@
   :url "http://github.com/threatgrid/ctim"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :pedantic? :abort
+  ;; :pedantic? :abort
```
ereteog, line 9: could you uncomment it please?

src/ctim/schemas/asset_mapping.cljc
```diff
@@ -26,13 +27,26 @@
...
-   (f/entry :stability Stability)
-   (f/entry :observable c/Observable)
-   (f/entry :asset_ref asset/AssetRef)
-   (f/entry :asset_type asset/AssetType)))
+   (f/entry :valid_time c/ValidTime
+            :description (str "For each asset, we allow for the assertion of time bound properties."
+                              "This gives us both a record of the current state of the asset,"
+                              "as well as history."))
+   (f/entry :confidence v/HighMedLow
+            :description "Is it susceptible to manipulation or translation.")
```
ereteog, line 35: I would be more general like in [Incident](https://github.com/threatgrid/ctim/blob/f9e2eff38e51ca0c2b8c4099457a1ecf07c8dba3/src/ctim/schemas/incident.cljc#L35) and you could use your description as an example):
"level of confidence held in the characterization of this AssetMapping (e.g. is it susceptible to manipulation or translation?)."

src/ctim/schemas/asset_mapping.cljc
```diff
@@ -26,13 +27,26 @@
...
+   (f/entry :specificity Specificity
+            :description (str "Denotes the level of how many assets "
+                              "potentially could have this same identifier."))
+   (f/entry :stability Stability
+            :description (str "Do we manage when it changes, "
+                              "or it always a time bound assignment?"))
```
ereteog, line 41:
```suggestion
                              "or is it always a time bound assignment?"))
```

src/ctim/schemas/asset_properties.cljc
```diff
@@ -25,8 +32,10 @@
   c/sourced-object-entries
   (f/required-entries
    (f/entry :type AssetPropertiesTypeIdentifier)
-   (f/entry :valid_time c/ValidTime)
-   (f/entry :asset_ref asset/AssetRef))
+   (f/entry :valid_time c/ValidTime
+            :description "The time range during which the AssetProperties considered valid.")
```
ereteog, line 36:
```suggestion
            :description "The time range during which the AssetProperties is considered valid.")
```

src/ctim/schemas/target_record.cljc
```diff
@@ -18,9 +18,14 @@
   "Schema for TargetRecord Targets"
   (:entries c/IdentitySpecification)
   (f/optional-entries
-   (f/entry :os f/any-str)
-   (f/entry :internal (f/bool :default false))
-   (f/entry :sensor f/any-str)
+   (f/entry :os f/any-str
+            :description (str "Source Operating System where TargetRecord was originated."))
+   (f/entry :internal (f/bool :default false)
+            :description "Is it internal to our network.")
```
ereteog, line 24:
```suggestion
            :description "Is it internal to our network?")
```

ereteog (approved):
LGTM

project.clj
agzam: Ah shoot... Good catch... Sorry about that.

PR #330, opened 2021-02-12
Bump `ctim-schema-version` to 1.1.3

halcyon:
- Bump `ctim-schema-version` to 1.1.3 
- fix malware schema: `describable` -> `described`.


ereteog (approved):
LGTM
Thank you for also fixing my error.
Could you just add a short description?

Issue #334, opened 2021-03-11
Bundle-maximal example should take into account asset-refs


agzam:
There is a specific association of a given `AssetMapping/AssetProperties` entity to some Asset, which is set by defining `:asset-ref` field of `AssetMapping/AssetProperties`

Bundle example should take that relation into account. The example should include Asset(s) with transient IDs and `AssetMapping/AssetProperties` entity(s) with `:asset-ref` with the value of the same transient IDs.

agzam:
This is a blocker for delivering https://github.com/threatgrid/ctia/pull/1089
Without it, the tests such as `ctia.bundle.routes-test/bundle-import-test` won't pass for the code that updates the Bulk Import logic described in the PR.

PR #335, opened 2021-03-15
Bundle example with asset-refs

agzam:
Closes: https://github.com/threatgrid/ctim/issues/334

No QA required.

### DIFF
--- a/src/ctim/examples/bundles.cljc
+++ b/src/ctim/examples/bundles.cljc
@@ -26,6 +26,15 @@
 (defn set-of [model]
   (set (repeat 3 model)))
 
+(def ^:private asset
+  (set-of (assoc asset-maximal :id "transient:asset-1")))
+
+(def ^:private asset-mapping
+  (set-of (assoc asset-mapping-maximal :asset_ref "transient:asset-1")))
+
+(def ^:private asset-properties
+  (set-of (assoc asset-properties-maximal :asset_ref "transient:asset-1")))
+
 (def bundle-maximal
   {:id                      "http://ex.tld/ctia/bundle/bundle-5023697b-3857-4652-9b53-ccda297f9c3e"
    :schema_version          c/ctim-schema-version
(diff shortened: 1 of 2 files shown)


ereteog (commented):
The problem to use transient ids in the `asset_ref` field for [`asset-properties`](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/asset_properties.cljc#L37) and [`asset-mapping`](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/asset_mapping.cljc#L48) is only the schema of this field which is currently [`ref-for-type`](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/asset.cljc#L47) that basically [expects a string url](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/common.cljc#L59).
In order to enable us to use a transient id for `asset_ref` field, you must replace it by a [`c/Reference`](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/common.cljc#L46) that accepts transient id like [`source_ref` and `target_ref` of relationship](https://github.com/threatgrid/ctim/blob/9fff33b81c705c649ad8ea8d9331fa091102f121/src/ctim/schemas/relationship.cljc#L19).

agzam:
I did this: https://github.com/threatgrid/ctim/pull/335/commits/f6f80289b36e6992e91113208a68284f52203e7e

Do you think I should revert my changes around `bundle-maximal`? Leave it with the long IDs, as it was?
I think it'd be nice to have a "proper" bundle example with transient IDs. No?

ereteog (commented):
> I did this: f6f8028

Nice 

> Do you think I should revert my changes around bundle-maximal? Leave it with the long IDs, as it was?
I think it'd be nice to have a "proper" bundle example with transient IDs. No?

Yes please rever it. the `*_refs` fields are intented to lighten bundle responses with only the ids. Thus it is mainly long ids that are expected, since they contain those from the response.

agzam:
> Yes please revert it.

I have reverted all the changes in Bundle examples. Now I need to bump the version, merge it and deploy new version.

ereteog:
LGTM

ereteog (approved):
LGTM

PR #343, opened 2021-05-13
Document modeling complex events

saintx:
Documenting how to model arbitrarily complex event data using indicators and sightings.


saintx:
Rolling this into another PR

PR #347, opened 2021-06-17
Incident entity description clarifications

saintx:
This update clarifies the definition and purpose of the incident entity,
specifically distinguishing the concepts of "incident" and "event", as
defined in NIST Special Publication 800-61, Revision 2, on Computer Security
Incident Handling, per discussions with Eric Hulse. The goal in making these
distinctions explicit in our schema is to clarify future model mappings for
complex adverse events using the `sighting` entity, and to give guidance for
continued development of the `incident` entity, which is explicitly for the
purpose of modeling *incident response* data.

Also includes field descriptions for `IncidentTime` inline fields, to help
clarify their intent and purpose.

[skip ci]

### DIFF
--- a/src/ctim/schemas/incident.cljc
+++ b/src/ctim/schemas/incident.cljc
@@ -8,21 +8,33 @@
 (def-map-type IncidentTime
   (concat
    (f/required-entries
-    (f/entry :opened c/Time))
+    (f/entry :opened c/Time
+             :description "Time the incident was first opened."))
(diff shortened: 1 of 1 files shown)


jbouwman (approved):
It's  helpful to have more fields described here as there have been many questions from UX and other CTIA consumers about semantics.

egierach (approved):

ereteog (approved):

PR #349, opened 2021-07-20
Add recently joined team members as code owners

rplevy:


rplevy:
@gbuisson it looks like @bartuka does not yet have access to this repo by the way

agzam (approved):

irinarenteria (approved):
LGTM, with the caveat that it's ok to protect certain paths if there's anything extra-sensitive.

rplevy:
To record the decision that was arrived at in a meeting we had about this, we're going to leave this open while we transition to protecting certain paths. This is probably all for @ereteog to decide as he has worked on it the most. I am assigning it to him.

ereteog:
The sensitivity is in CTIA, when bumping the CTIM version: if a field is added and the mapping not updated accordingly, it will not be indexed and further fixes will require a migration to index previously inserted values. 
So anyone that hit the "approve" button must consider the integration in CTIA. Never forget that migrating indices like the sighting store with hundreds of millions of documents is a very expensive operation since it contains

PR #350, opened 2021-08-05
Adds serial_number to observable vocabulary

agzam:
Closes: https://github.com/advthreat/iroh/issues/5559


gbuisson (approved):
LGTM

Before merging:

- generate the doc and commit `lein doc`

gbuisson:
This was supposed to be "computer_serial_number" though but the ratio work/benefit to change that again might not be worth it at this point in time.

agzam:
I've discussed the change with Jyoti, she instructed me to name it like that so we can unlock the other team sooner, without forcing them to change their stuff.

PR #352, opened 2021-08-10
fix bundle example

ereteog:
fix the bundle example (the transient id of the sighting in a relationships was not equal to the id of the sighting)


agzam (approved):
Good!

PR #353, opened 2021-08-30
Add Enrichment Status Field

halcyon:
Progress toward https://github.com/threatgrid/response/issues/842

### DIFF
--- a/doc/json/bundle.json
+++ b/doc/json/bundle.json
@@ -46,6 +46,11 @@
       "rejected" : "2016-01-01T01:01:01.000Z"
     },
     "discovery_method" : "string",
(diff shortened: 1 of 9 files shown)


ereteog (commented):
Before Merging it, I would like to discuss the design of this state.
Here we are introducing the tracking of an IROH process in a generic model of incident.
For instance we removed Investigation from CTIM because it was an object representing the state of the engine process.
If the final result will be to have an Investigation object linked to an incident, what about storing this state in Investigation and creating a relationship between the incident and corresponding investigation?

halcyon:
Closing in favor of https://github.com/advthreat/iroh/issues/5710

PR #364, opened 2022-02-18
move to actions

frenchy64:
Travis build can be disabled after this has been merged. We'll add a protected branch check for the `all-pr-checks` Actions job and remove Travis.

Some generator deps are bumped in the hope it plays nicer with Java 17. The build is twice as slow for some reason, so I've moved it to scheduled builds only (not PR's and pushes).


agzam (approved):
Nice!

ereteog (approved):
thx!

PR #368, opened 2022-07-06
Issue 336: Adding Kenna Risk Scores to Mitre Attack Patterns

michaels-den:
⚠️ Seeking feedback: are there any other places within CTIM that require change for an optional feed addition?

This work adds an optional field of `risk_score` to the
`attack_pattern` schema. Possible values are PosInts.

Resolves #336


ereteog:
I designed it otherwise. The data will be in IROH for now.

PR #370, opened 2022-08-11
Adding new relationship types to the vocabulary for Techniques and Sub-Techniques

jyoverma:
Adding new relationship types to the vocabulary for Techniques and Sub-Techniques

### DIFF
--- a/src/ctim/schemas/vocabularies.cljc
+++ b/src/ctim/schemas/vocabularies.cljc
@@ -427,7 +427,9 @@
     "sighting-of"
     "targets"
     "uses"
-    "variant-of"})
+    "variant-of"
+    "technique-of"
+    "subtechnique-of})
 
 (def-enum-type RelationshipType
   relationship-type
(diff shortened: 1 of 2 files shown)


src/ctim/schemas/vocabularies.cljc
```diff
@@ -427,7 +427,9 @@
     "sighting-of"
     "targets"
     "uses"
-    "variant-of"})
+    "variant-of"
+    "technique-of"
```
msprunck, line 431: What is the purpose of the `technique-of` relation type? Between what type of entities would it be used?

Could you also update the documentation https://github.com/threatgrid/ctim/blob/master/doc/defined_relationships.md that defines how to use relation types?

frenchy64:
This PR addresses the feedback here (I forgot to check if this PR existed) https://github.com/threatgrid/ctim/pull/372

frenchy64:
Fixed by https://github.com/threatgrid/ctim/pull/372

PR #371, opened 2022-08-12
Add tactics/techniques field to incident

frenchy64:
https://github.com/advthreat/iroh/issues/6972

We want ES to index these fields for sorting/filtering purposes. I think we need to run the ES update-mappings task after we bump CTIA.

Used in https://github.com/threatgrid/ctia/pull/1294


marioaquino (approved):
LGTM

PR #392, opened 2023-01-17
restricts score generator

agzam:
Incident scores can't have NaN and Infinite values, and sometimes they get generated

### DIFF
--- a/src/ctim/lib/generators.clj
+++ b/src/ctim/lib/generators.clj
@@ -72,6 +72,11 @@
    string/join
    (gen/vector open-vocab-char 1 100)))
 
+(def incident-score
+  (gen/double*
+   {:infinite? false
+    :NaN? false}))
+
 ;; vulnerability
 
 (def score
--- a/src/ctim/lib/generators.clj
+++ b/src/ctim/lib/generators.clj
@@ -75,11 +75,7 @@
 ;; vulnerability
 
 (def score
-  (gen/double*
-   {:infinite? false
-    :NaN? false
-    :min 0.0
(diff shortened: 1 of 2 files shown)


agzam:
I'm thinking, maybe I should exclude negative values as well, or it's better if it remains a bit unrestrained, maybe for some weird testing scenarios?

ereteog (commented):
can you move the generator with [other ones](https://github.com/threatgrid/ctim/blob/master/src/ctim/lib/generators.clj), please?

agzam:
> can you move the generator with [other ones](https://github.com/threatgrid/ctim/blob/master/src/ctim/lib/generators.clj), please?

Initially, that's what I wanted to do, in the first commit I used `vulnerability.score` generator. Turns out, `Incident.score` and that one are not the same things. They aren't, right?

ereteog:
> > can you move the generator with [other ones](https://github.com/threatgrid/ctim/blob/master/src/ctim/lib/generators.clj), please?
> 
> Initially, that's what I wanted to do, in the first commit I used `vulnerability.score` generator. Turns out, `Incident.score` and that one are not the same things. They aren't, right?

They are not, but new definition should be with other ones in https://github.com/threatgrid/ctim/blob/master/src/ctim/lib/generators.clj or [the generators are all centralized (like gen/double) ](https://github.com/threatgrid/ctim/tree/master/src/ctim/generators)

@msprunck @gbuisson what's the best place to put new generators?

agzam:
appears to be fixed in https://github.com/threatgrid/ctim/pull/390

PR #398, opened 2023-02-21
Change incident score structure

frenchy64:
Related: https://github.com/threatgrid/ctia/pull/1340

The current `scores` structure requires convoluted ES features to fully support. We decided this to avoid migrations every time we add a new score type. But now we're going to change `scores` to be more intuitive because we will use [dynamic mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-mapping.html) to maintain flexibility of adding score types.

This will require a migration (of some kind) on INT. But in exchange it's much simpler to implement this new shape in CTIA.

Used an `enum` for the score types to help CTIA generate GraphQL objects (I didn't dig into why `f/any` or `f/str` didn't work).

### DIFF
--- a/doc/json/bundle.json
+++ b/doc/json/bundle.json
@@ -69,7 +69,7 @@
     "description" : "string",
     "external_ids" : [ "string" ],
     "scores" : {
-      "string" : 10.0
+      "keyword" : 10.0
     }
   } ],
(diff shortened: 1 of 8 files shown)


ereteog (approved):
Nice!

michaels-den (approved):
Looks good to me

PR #401, opened 2023-02-24
Added related_entities, note_class to Note

gbuisson:
#close https://github.com/advthreat/iroh/pull/7569

Add new fields to the Note Entity as a mean to manage Incident Task COAs progression.

### DIFF
--- a/src/ctim/schemas/note.cljc
+++ b/src/ctim/schemas/note.cljc
@@ -1,7 +1,7 @@
 (ns ctim.schemas.note
   (:require [ctim.schemas.common :as c]
             [ctim.schemas.vocabularies :as v]
-            #?(:clj  [flanders.core :as f :refer [def-entity-type def-enum-type def-eq]]
+            #?(:clj  [flanders.core :as f :refer [def-entity-type def-map-type def-eq]]
                :cljs [flanders.core :as f :refer-macros [def-entity-type def-eq]])))
 
 (def type-identifier "note")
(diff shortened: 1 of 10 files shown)


src/ctim/schemas/note.cljc
```diff
@@ -20,6 +21,8 @@ For example, an analyst may add a Note to a Campaign object created by another o
    (f/entry :entity_id c/Reference)
    (f/entry :content c/Markdown))
   (f/optional-entries
+   (f/entry :note_class v/NoteClass)
+   (f/entry :related_entities (f/seq-of c/Reference))
```
ereteog, line 25: I would use `note_type` in order to be consistent with `relationship_type` of [relationship schema](https://github.com/threatgrid/ctim/blob/4aa3a14db97715c493a05dec581984b2ae542af5/src/ctim/schemas/relationship.cljc#L18).
I would also use `entity_ids` like `entity_id` of [feedback](https://github.com/threatgrid/ctim/blob/4aa3a14db97715c493a05dec581984b2ae542af5/src/ctim/schemas/feedback.cljc#L17) and `asset_id` of asset-mappings and asset-properties.

gbuisson: I can't use NoteTypenas it conflicts with another eponym field in the vocabularies

gbuisson: I debates the use of entity_ids but I wanted perhaps to convey the notion of a primary target for the relation

gbuisson: we finally decided to switch to `related_entities` as a single vector of maps including the target type for convenience.

src/ctim/schemas/vocabularies.cljc
```diff
@@ -1509,3 +1509,16 @@
 (def-enum-type PromotionMethod promotion-method
   :open? false
   :gen (cs/gen promotion-method))
+
+(def note-class
```
ereteog, line 1513: if we proxify it, we could make it open and restrict values in the proxy

ereteog (approved):
LGTM

src/ctim/schemas/vocabularies.cljc
gbuisson: that's a good point

gbuisson: done

PR #387, opened 2022-11-30
Adding endpoint event ontologies

jyoverma:
- Added event ontologies
- Added context to Sighting schema
- Added Sighting examples for new event types

PR #405, opened 2023-03-07
Revert endpoint event ontologies

msprunck:
Design https://github.com/advthreat/iroh/pull/7167

This PR reverts the event ontologies added by #387 because they causing test failures in CTIA. We keep the new observable types as they are required by the SCA integration which has a higher priority. It will give us more time to fix the test failures in CTIA.


frenchy64 (approved):
Code looks good, could you link an issue explaining the rationale?

src/ctim/schemas/vocabularies.cljc
```diff
@@ -259,7 +259,7 @@
     "cisco_mid"
     "cisco_uc_id"          ;; Cisco Unified Connector ID
     "cisco_cm_id"          ;; Cisco NVM UUID
-    "device"               
+    "device"
```
jyoverma, line 262: we need the process observables here

jyoverma (changes requested):
other than adding back the observable type fields for process, the PR looks good

src/ctim/schemas/vocabularies.cljc
msprunck: They are still there, I did not remove them.

msprunck:
> Code looks good, could you link an issue explaining the rationale?

I updated the PR description with a link to the design doc and I explained the approach. I hope it clarifies the intent.

src/ctim/schemas/vocabularies.cljc
gbuisson: @jyoverma I think you got confused, when you check a PR on github, the interface is showing you what changed only and not the full source files every time, hence the observables are still here, if you want to make sure select "view file" in the interface.

jyoverma: Ok in that case, I approve these changes

jyoverma (approved):
Looks good

PR #407, opened 2023-03-10
Restore Sighting contextual events

msprunck:
Connected to https://github.com/threatgrid/ctim/pull/405

Restored sighting contextual events to work on their integration in CTIA. Some tests were failing in CTIA.


jyoverma (approved):
Looks good!

PR #411, opened 2023-03-27
Contextual events fixes

msprunck:
- Use `any-int` for Traffic.protocol which is the [IP protocol number](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml)
- Remove `:creation_time` from `ProcessCreateEvent`s
- Set `HTTPEvent.url_port` optional


jyoverma (approved):
Looks good to me. Thanks!

PR #416, opened 2023-05-11
adds incident_time.created

agzam:
We need to expose `created` field even though it won't be stored explicitly in CTIA.

> Related: https://github.com/advthreat/iroh/issues/7866

The discussion: https://github.com/advthreat/iroh/issues/7853#issuecomment-1540591124

### DIFF
--- a/doc/json/incident.json
+++ b/doc/json/incident.json
@@ -1,43 +1,58 @@
 {
-  "confidence" : "string",
-  "tlp" : "string",
-  "techniques" : [ "string" ],
-  "id" : "string",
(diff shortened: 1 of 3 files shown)


ereteog (changes requested):
we did not concluded that. Wait for the team.

agzam:
> we did not concluded that. Wait for the team.

Of course. Even if approved, I have to run a bunch of tests, etc. by the way. for some reason it's allowing me to merge even without any reviews and with a change request for some reason. Can you check the repo settings?

image

ereteog:
see https://github.com/advthreat/iroh/issues/7917

PR #417, opened 2023-05-12
Adds Playbooks for COA composition

michaels-den:
⚠️ Work in Progress: Opening to begin discussion. ⚠️ 

**CTIM addition of Playbooks for composable response actions**

This work adds an ability to compose collections of COAs by their phase (also known as attack stage) in the form of a Playbook. The intention is to create generally available Playbooks, composed of generally available COAs that can drive Playbook coordinated response. Private playbooks can also be generated.


ereteog:
I think that we can close this one, Devin developed the API.
wdyt?

PR #421, opened 2023-06-21
Documentation updates

agzam:
Improved and expanded descriptions of several fields.

_You can largely ignore changes in .markdown files, since it's generated content and repeats the same text for fields shared across entities._


src/ctim/schemas/incident.cljc
```diff
@@ -59,7 +59,18 @@
...
+             "Field is used to indicate the severity or impact of the threat represented by the incident."
+             "\n\n"
+             "- `asset` - assesses the potential damage or harm that the threat can cause to the affected asset or assets. "
+             "The scale ranges from 0 to 10, with 10 indicating the highest potential harm. "
+             "\n"
+             "`global` - assesses the overall impact or significance of the threat to the organization or wider "
+             "community. The scale ranges from 0 to 1000, with 1000 indicating the highest impact."
+             "\n"
+             "`tlp` - assesses the sophistication or complexity of the attack techniques used by the threat actor. "
+             "The scale ranges from 0 to 100, with 100 indicating the highest level of sophistication"))))
```
ereteog, line 73: that's supposed to be generic from the CTIM perspective. asset, global, ttp are configured in CTIA and should not be mentioned here.

agzam: Maybe changing the language would improve it? 
How about explicitly saying that these are example values?

agzam: Check "explicitly note that score types are examples" commit. wdyt?

ereteog: it still leak an internal logic that we intented to keep private

agzam: It's not really "private", the stuff is right there in our schema: https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/incident.cljc#L40

This was the actual reason why I started updating documentation because I couldn't find an explanation for these things anywhere. I think we have to try to explain them at least, if not here, then somewhere else. CTIM seems like the right place for it.

PR #428, opened 2023-10-18
change select sighting context fields to optional

t2sw:
Per 8427 make the following changes to sighting context

Make [HTTPType > :method](https://github.com/threatgrid/ctim/blob/f6529ae7f56964c41fb69a0db06b3aa5465b815b/src/ctim/schemas/sighting/context.cljc#L163C33-L163C33) optional
Make RegistryDeleteType  > :registry_value optional
Make [RegistrySetType > :registry_data](https://github.com/threatgrid/ctim/blob/f6529ae7f56964c41fb69a0db06b3aa5465b815b/src/ctim/schemas/sighting/context.cljc#L196) optional

### DIFF
--- a/doc/structures/bundle.md
+++ b/doc/structures/bundle.md
@@ -7812,12 +7812,11 @@ Time of the observation. If the observation was made over a period of time, than
 |[process_id](#propertyprocess_id-integer)|Integer| |&#10003;|
 |[process_name](#propertyprocess_name-shortstringstring)|ShortStringString| |&#10003;|
 |[registry_key](#propertyregistry_key-shortstringstring)|ShortStringString| |&#10003;|
-|[registry_value](#propertyregistry_value-medstringstring)|MedStringString| |&#10003;|
 |[time](#propertytime-observedtimeobject)|*ObservedTime* Object| |&#10003;|
 |[type](#propertytype-registrydeletetypeidentifierstring)|RegistryDeleteTypeIdentifierString| |&#10003;|
 |[process_guid](#propertyprocess_guid-integer)|Integer| ||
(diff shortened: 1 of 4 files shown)


msprunck (dismissed):
LGTM!

samwagg:
Do we need to make the same changes to casebook?

t2sw:
@samwagg @msprunck i will make the corresponding changes for casebook

msprunck:
> @samwagg @msprunck i will make the corresponding changes for casebook

Could you explain how this change can also be applied to Casebooks?

samwagg:
> > @samwagg @msprunck i will make the corresponding changes for casebook
> 
> Could you explain how this change can also be applied to Casebooks?

I'm not very familiar with Casebook. I just noticed that casebook.md also has this HTTPType, and the doc wasn't updated in this PR. Looking closer, it seems that Casebook.md includes the `sighting` schema. So I assume the doc just needs to be regenerated from the same schema?

But I am totally open to the idea that I'm just confused :)

samwagg:
> > > @samwagg @msprunck i will make the corresponding changes for casebook
> > 
> > 
> > Could you explain how this change can also be applied to Casebooks?
> 
> I'm not very familiar with Casebook. I just noticed that casebook.md also has this HTTPType, and the doc wasn't updated in this PR. Looking closer, it seems that Casebook.md includes the `sighting` schema. So I assume the doc just needs to be regenerated from the same schema?
> 
> But I am totally open to the idea that I'm just confused :)

I suppose the same is true of Bundle. Are there others?

samwagg:
Just Bundle and Casebook. I pulled down the branch and ran `lein doc` to confirm.

msprunck (approved):
LGTM!

PR #433, opened 2023-11-14
Incident meta field

ereteog:
Applying @gbuisson feedback on the team chan:
> Guillaume Ereteo you forgot to wait for the reviews before merging, the schema change you propose for CTIM should be more constrained, I think the keys of the map should be enforced as keyword and the values as scalar types only.


gbuisson (approved):
SGTM

PR #439, opened 2024-01-15
time metadata

ereteog:
> Related #advthreat/iroh/issues/8788

In order to (try to) end the confusion with timestamp in all documents that is dedicated to date control by sources and incident_time various fields like opened that represent status changes, we decided to expose the CTIA internal document time meta: `created` and `modified`.
This will clarify:
- what is displayed in the UI
- the understanding of the `[from, to[` filtering that is based on created
- ease the QA, that needs a DB access to see these fields.

### DIFF
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -184,7 +184,7 @@
                     "material.")
   :reference "[External Reference](https://docs.google.com/document/d/1dIrh1Lp3KAjEMm8o2VzAmuV0Peu-jt9aAh1IHrjAroM/pub#h.72bcfr3t79jx)")
 
-(def base-entity-entries
+(def base-editable-entity-entries
   (concat
    (f/required-entries
     (f/entry :id ID
@@ -224,6 +224,8 @@
                                "information that may not have a standardized format or identifier."))
     (f/entry :timestamp Time
              :description "The time this object was created at, or last modified.")
+    (f/entry :timestamp Time
+             :description "The time this object was created at, or last modified.")
     (f/entry :language ShortString
              :description (str "The `language` field is used to specify the primary language of the affected "
                                "system or the target of an attack. It can be used to provide additional context "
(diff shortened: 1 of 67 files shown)


src/ctim/schemas/common.cljc
```diff
@@ -255,10 +257,21 @@
...
 
+(def time-metas
+   (f/optional-entries
+    (f/entry :created Time
+             :description "The time this object was created at.")
+    (f/entry :modified Time
+             :description "The time this object was last modified.")))
+
+(def base-entity-entries
+  "Base for New Entities, optionalizes ID and type and schema_version"
```
msprunck, line 268: The docstring looks wrong

src/ctim/schemas/common.cljc
```diff
@@ -224,6 +224,8 @@
                                "information that may not have a standardized format or identifier."))
     (f/entry :timestamp Time
              :description "The time this object was created at, or last modified.")
+    (f/entry :timestamp Time
+             :description "The time this object was created at, or last modified.")
```
msprunck, line 228: Duplicate entry

src/ctim/schemas/common.cljc
```diff
@@ -255,10 +257,21 @@
                                "general information about security threats might be marked as `amber` or "
                                "`green`, indicating that it can be shared more broadly within an organization.")))))
 
+(def time-metas
+   (f/optional-entries
+    (f/entry :created Time
+             :description "The time this object was created at.")
+    (f/entry :modified Time
+             :description "The time this object was last modified.")))
```
msprunck, line 265: `timestamp` can be set in a new entity but not `created`. For a CTIM user, it is not clear what is the difference. Could you clarify the usage in the description? Do we expect to have the same value in `timestamp` and `created` when the entity is new and the same value in `timestamp` and `modified` when it was updated?

ereteog: I updated the doc.
We do set `timestamp` with the value of `created` for incidents when it's not provided.
If it's provided by the source, we keep the provided value, and we update it at each modification in the [default realize](https://github.com/threatgrid/ctia/blob/7bdefcc06f7ec7e74575ed841a92d36ae4a2c1af/src/ctia/domain/entities.clj#L38).
However, since this [PR](https://github.com/threatgrid/ctia/pull/1377) we have a custom realize for incident that do not update timestamp on modification. We [decided](https://github.com/advthreat/iroh/issues/7853#issuecomment-1562811852) that after another confusion, but we did not want to apply it to all entiites (IMO it should be).
So in this context, I hope that my update to the doc still help users of CTIM.
Do you have another proposal?

msprunck: If `created` and `modified` are specific to the application in charge of the persistence, I would only change the schema and expose it in the application.

ereteog: The purpose is to enable it to be consumed by users of the application.
Only the application can modify it, but the information that convey these data can be consumed by API clients and is not specific to the application.
In particular, [we can display it in the UI](https://github.com/advthreat/iroh/issues/8788#issuecomment-1887314105).

ereteog:
will apply the same logic than groups and owner which schemas are already duplicated everywhere it's used.

PR #445, opened 2024-04-24
Add process_uid to the observable type vocabulary.

brookeswanson:
Sighting.relations fields often include self referential processes.

For example:

* cmd.exe `parent_of` cmd.exe
* args 1 `args_of` cmd.exe
* args 2 `args_of` cmd.exe
* cmd.exe `connected_to` hostname

This does not tell us a very clear picture because there is no way to determine which thing did what.

This will give us the option of having a unique id for a process which is a collection of the attributes of that process.


brookeswanson:
@jyoverma

michaels-den (approved):
I dig it - great add.

PR #449, opened 2024-07-08
Improve swagger docs

frenchy64:
https://github.com/advthreat/iroh/issues/8946

Via flanders changes, adds examples to the json docs and swagger metadata.

We reuse `:default` to populate examples here. This has undesirable effects on the documentation. I reviewed the existing documentation around defaults, and many were not appropriate for user-facing docs. The closest to useful was that tlp defaults to green, but this is configurable in CTIA so that is debatable.

Instead of developing another flanders feature, I disabled default values from printing in the docs. We might want to revisit this if we actually use defaults in a user-facing way (not just generating examples).


yogsototh:
For the build perhaps you could deploy a SNAPSHOT version in clojars and use this one?

src/ctim/schemas/common.cljc
```diff
@@ -377,8 +377,11 @@
 
 (def-map-type Observable
   (f/required-entries
-   (f/entry :value f/any-str)
-   (f/entry :type v/ObservableTypeIdentifier))
+   (f/entry :value f/any-str
+            :default "1.2.3.4"
+            :description "The value of the observable.")
+   (f/entry :type v/ObservableTypeIdentifier
```
yogsototh, line 383: As you provided a default value for the observable value that is an IPv4 address perhaps you can also set a default value for the observable type?

```suggestion
   (f/entry :type v/ObservableTypeIdentifier
            :default "ip"
```

test/ctim/test_helpers/core.cljc
```diff
@@ -2,7 +2,9 @@
   (:require [clojure.spec.alpha :as cs]
             [clojure.spec.test.alpha :as stest]
             [ctim.domain.id :as id]
-            [flanders.spec :as fs])
+            [flanders.spec :as fs]
+            [flanders.schema :as fschema]
+            )
```
yogsototh, line 7:
```suggestion
            [flanders.schema :as fschema])
```

yogsototh (approved):
LGTM!

ereteog (approved):

PR #453, opened 2024-11-11
Add acudid and cvm_id observable types

DeLaGuardo:


src/ctim/schemas/vocabularies.cljc
```diff
@@ -274,15 +274,17 @@
   :gen (cs/gen motivation))
 
 (def observable-type-identifier
-  #{"amp_computer_guid"
+  #{"acudid"
```
msprunck, line 277:
```suggestion
  #{"acudid" ;; AnyConnect ID
```

msprunck (approved):
LGTM!

PR #461, opened 2025-02-25
add new Device Insights observables

frenchy64:
https://cisco-sbg.atlassian.net/browse/XDR-21205

Also updated the build to be much faster thanks to https://github.com/DeLaGuardo/setup-clojure/pull/109


msprunck (approved):
LGTM!

PR #465, opened 2025-05-21
XDR-27079 Add incident_time contained.

samwagg:
Needed to support a new interval metric in CTIA: new_to_contained


ereteog (approved):
LGTM

PR #483, opened 2026-03-12
Add size constraints to CTIM collection fields (XDR-46972)

ereteog:
> **Epic** [XDR-46972](https://cisco-sbg.atlassian.net/browse/XDR-46972)
> Related [flanders#58](https://github.com/threatgrid/flanders/pull/58)
> Related [ctia#1514](https://github.com/threatgrid/ctia/pull/1514)

Add `pred/max-len` constraints to collection fields in sighting, incident, and common schemas.

### Changes

- Bump flanders to `1.1.1-SNAPSHOT` (adds `:spec` support on `SequenceOfType`, `SetOfType`, and `MapType`)
- Add `default-collection-max-len` (500) constant in `common.cljc` for shared collection fields
- Constrain `external_ids` and `external_references` in base entity entries
- Add specific limits for sighting fields:
  - `columns`: 100, `rows`: 10,000
  - `targets`: 2,000, `observables`: 5,000, `relations`: 10,000
- Constrain incident fields (`categories`, `assignees`, `tactics`, `techniques`) to 500
- Add `relation_info.actions` limit of 1,000
- Add `{:max-len len}` metadata to `pred/max-len` for downstream 413 error reporting
- Replace unbounded `f/any-str` with typed string constraints (`ShortString`, `MedString`, `LongString`)

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed. All 139 existing tests (including generative tests) pass.

<a name="release-notes">[§](#release-notes)</a> Release Notes
=============================================================

```
intern: Add size constraints to CTIM collection fields.
```

<a name="squashed-commits">[§](#squashed-commits)</a> Squashed Commits
======================================================================

[XDR-46972]: https://cisco-sbg.atlassian.net/browse/XDR-46972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,7 @@
                  [com.google.protobuf/protobuf-java "4.32.1"] ;clj-momo > org.clojure/clojurescript
                  [threatgrid/clj-momo "0.4.1"]
                  [org.mozilla/rhino "1.8.0"] ;threatgrid/flanders > kovacnica/clojure.network.ip
-                 [threatgrid/flanders "1.1.0"]
+                 [threatgrid/flanders "1.1.1-SNAPSHOT"]
                  [metosin/ring-swagger "1.0.0"]
                  [org.clojure/test.check "1.1.1"]
                  [com.gfredericks/test.chuck "0.2.15"]
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -24,6 +24,9 @@
 
 (def ctim-schema-version (ctim-version))
 
+;; Default size limit for collection fields (strings, references, etc.)
+(def default-collection-max-len 500)
+
 (def-eq CTIMSchemaVersion ctim-schema-version)
 
 (cs/def ::ctim-schema-version
@@ -197,7 +200,8 @@
    (f/optional-entries
     (f/entry :revision PosInt
              :description "A monotonically increasing revision, incremented each time the object is changed.")
-    (f/entry :external_ids f/any-string-seq
+    (f/entry :external_ids (f/seq-of f/any-str
+                                     :spec (pred/max-len default-collection-max-len))
              :description (str "It is used to store a list of external identifiers that can be linked to the "
                                "incident, providing a reliable and manageable way to correlate and group related "
                                "events across multiple data sources. It is especially useful in larger "
@@ -209,7 +213,8 @@
                                "information can be shared among incident management systems. It can be used to "
                                "cross-reference with other external tools such as threat intelligence feeds and "
                                "vulnerability scanners."))
-    (f/entry :external_references [ExternalReference]
+    (f/entry :external_references (f/seq-of ExternalReference
+                                            :spec (pred/max-len default-collection-max-len))
              :description (str "Specifies a list of external references which refers to non-CTIM "
                                "information.\n\n"
                                "Similar to `external_ids` field with major differences:\n\n"
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -27,6 +27,9 @@
 ;; Default size limit for collection fields (strings, references, etc.)
 (def default-collection-max-len 500)
 
+;; Max number of actions allowed in a single relation's relation_info
+(def max-relation-actions 1000)
+
 (def-eq CTIMSchemaVersion ctim-schema-version)
 
 (cs/def ::ctim-schema-version
@@ -628,7 +631,12 @@
             :required? false)
    (f/entry :relation ObservableRelationType)
    (f/entry :relation_info (f/map
-                            [(f/entry f/any-keyword f/any)])
+                            [(f/entry f/any-keyword f/any)]
+                            :spec (fn [m]
+                                    (let [actions (:actions m)]
+                                      (or (nil? actions)
+                                          (not (sequential? actions))
+                                          ((pred/max-len max-relation-actions) actions)))))
             :required? false)
    (f/entry :source Observable)
    (f/entry :related Observable)]
--- a/src/ctim/schemas/common.cljc
+++ b/src/ctim/schemas/common.cljc
@@ -178,9 +178,9 @@
     (f/entry :description Markdown)
     (f/entry :url URI
              :description "A URL reference to an external resource.")
-    (f/entry :hashes f/any-string-seq
+    (f/entry :hashes (f/seq-of ShortString)
              :description "Specifies a dictionary of hashes for the contents of the url.")
-    (f/entry :external_id f/any-str
+    (f/entry :external_id ShortString
              :description "An identifier for the external reference content.")))
   :description (str "External references are used to describe pointers to information "
                     "represented outside of CTIM. For example, a Malware object could "
@@ -194,13 +194,13 @@
    (f/required-entries
     (f/entry :id ID
              :description "Globally unique URI identifying this object.")
-    (f/entry :type f/any-str)
+    (f/entry :type ShortString)
     (f/entry :schema_version SchemaVersion
              :description "CTIM schema version for this entity."))
    (f/optional-entries
     (f/entry :revision PosInt
              :description "A monotonically increasing revision, incremented each time the object is changed.")
-    (f/entry :external_ids (f/seq-of f/any-str
+    (f/entry :external_ids (f/seq-of ShortString
                                      :spec (pred/max-len default-collection-max-len))
              :description (str "It is used to store a list of external identifiers that can be linked to the "
                                "incident, providing a reliable and manageable way to correlate and group related "
@@ -268,7 +268,7 @@
    base-entity-entries
    (f/optional-entries
     (f/entry :id ID)
-    (f/entry :type f/any-str
+    (f/entry :type ShortString
              :description "A valid entity type identifer")
     (f/entry :schema_version CTIMSchemaVersion
              :description "CTIM schema version for this entity."))))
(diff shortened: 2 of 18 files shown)


yogsototh (approved):
I have only two concerns:

- why these numbers, to me they are a bit like "magic numbers" and it will eventually breaks in PROD refusing the creation of many entities suddenly without a warning which may create issues with already working integrations. Personally, I think it may be a bit too much to constrain statically here instead of directly into the CTIA application where we could apply different size constraints depending on the source for example. Which will give the time for an integration to adapt to the new constraints. And perhaps instead of a hard reject, we may also integrate a mechanism to just cut the strings/arrays to keep the first one.

- We should check if adding these constraints does not affect Swagger and Swagger UI. From memory plumatic/schemas and swagger do not support very well certain kind of external constraints like this one. So while the API will work, Swagger UI may be broken, or not show the correct schemas anymore.

project.clj
```diff
@@ -12,7 +12,7 @@
                  [com.google.protobuf/protobuf-java "4.32.1"] ;clj-momo > org.clojure/clojurescript
                  [threatgrid/clj-momo "0.4.1"]
                  [org.mozilla/rhino "1.8.0"] ;threatgrid/flanders > kovacnica/clojure.network.ip
-                 [threatgrid/flanders "1.1.0"]
+                 [threatgrid/flanders "1.1.1-SNAPSHOT"]
```
yogsototh, line 15: **CRITICAL** -- **SNAPSHOT dependency on unreleased flanders**: `1.1.1-SNAPSHOT` has no flanders git tag, no GitHub release, and Clojars 404s for it. The `:spec` support on `SequenceOfType`/`SetOfType`/`MapType` that this entire PR depends on exists only on flanders `master`, not in 1.1.0. Merging this to `master` makes the build non-reproducible (a SNAPSHOT is mutable and can be GC'd), blocks any stable ctim release, and leaves these constraints inert until flanders 1.1.1 ships. Release flanders 1.1.1 to Clojars and pin the released version before merge.

src/ctim/schemas/indicator.cljc
```diff
@@ -28,33 +28,33 @@
...
+   (f/entry :SIOC c/MedString))
   :description "An indicator which runs in snort...")
 
 (def-eq OpenIOCSpecificationType "OpenIOC")
 
 (def-map-type OpenIOCSpecification
   (f/required-entries
    (f/entry :type OpenIOCSpecificationType)
-   (f/entry :open_IOC f/any-str))
+   (f/entry :open_IOC c/MedString))
```
yogsototh, line 57: **HIGH** -- **Bound likely too small for documented content**: this field's docstring describes it as "an XML blob of an openIOC indicator," and OpenIOC XML documents routinely exceed 2048 chars, so `MedString` will reject most real `open_IOC` values. `SIOC` (line 49) and `snort_sig` (line 41) share the same 2048 cap and the same risk. More broadly, replacing unbounded `f/any-str` with a hard cap rejects documents that validated before, with no migration path. Recommend auditing existing PROD data against these bounds and confirming the values are intended for these blob fields.

src/ctim/schemas/common.cljc
```diff
@@ -618,12 +626,17 @@
...
    (f/entry :origin_uri URI
             :required? false)
    (f/entry :relation ObservableRelationType)
    (f/entry :relation_info (f/map
-                            [(f/entry f/any-keyword f/any)])
+                            [(f/entry f/any-keyword f/any)]
+                            :spec (fn [m]
+                                    (let [actions (:actions m)]
+                                      (or (nil? actions)
+                                          (not (sequential? actions))
```
yogsototh, line 638: **MEDIUM** -- **Actions cap is bypassable**: the limit only applies when `:actions` is `sequential?`. A non-sequential value -- a JSON object (decodes to a map) or a string -- makes `(not (sequential? actions))` return `true`, so it passes unbounded. Only the literal `:actions` key is inspected; the rest of this free-form `f/any` map is uncapped. If the goal is to bound oversized `relation_info`, consider rejecting non-sequential `:actions` rather than waving them through, and confirm `:actions` is the real key used in production payloads.

src/ctim/schemas/incident.cljc
```diff
@@ -127,13 +128,15 @@
...
             :description "A set of categories for this incident.")
    (f/entry :discovery_method v/DiscoveryMethod
             :description "Identifies how the incident was discovered.")
    (f/entry :intended_effect v/IntendedEffect
             :description "Specifies the suspected intended effect of this incident")
-   (f/entry :assignees [c/ShortString]
+   (f/entry :assignees (f/seq-of c/ShortString
+                                 :spec (pred/max-len c/default-collection-max-len))
             :description "A set of owners assigned to this incident.")
    (f/entry :detection_sources [c/MedString]
```
yogsototh, line 141: **MEDIUM** -- **Uncapped sibling**: `:detection_sources` kept the plain `[c/MedString]` form while its neighbours `:categories`, `:assignees`, `:tactics`, and `:techniques` all received `(f/seq-of ... :spec (pred/max-len default-collection-max-len))` in this PR. If the size cap is meant to cover incident collection fields, this one was missed; if it is intentionally unbounded, a one-line note would make that explicit. (Same asymmetry exists between the bounded embedded `SightingDataTable` and the unbounded standalone `DataTable` entity in `data_table.cljc`.)

yogsototh (changes requested):
**Verdict: CHANGES REQUESTED** -- 1 CRITICAL, 2 HIGH, 3 MEDIUM

This is a clean, well-scoped tightening and the core `max-len` predicate is correct (inclusive `>=` bound, nil-safe via `string?` guards). The blocking issues are about the dependency and where the constraints actually take effect, not the schema design itself.

### Findings

- **CRITICAL** [cross-cutting] -- **SNAPSHOT dependency on unreleased flanders**: `project.clj` pins `threatgrid/flanders "1.1.1-SNAPSHOT"`. flanders 1.1.1 has no git tag, no GitHub release, and Clojars returns 404 for it. The `:spec`-on-collection-types support that every `f/seq-of`/`f/map` constraint in this PR relies on exists only on flanders master, not in 1.1.0. Merging this to `master` makes the ctim build non-reproducible (a SNAPSHOT is mutable / can be GC'd), blocks cutting any stable ctim release, and means the collection constraints are inert until flanders 1.1.1 actually ships. Release flanders 1.1.1 to Clojars and pin the released coordinate before merge. (See inline on `project.clj`.)

- **HIGH** [cross-cutting] -- **Collection caps are not enforced in the schema/Swagger target**: flanders `schema.cljc` (master) does not read `:spec` for `SequenceOfType`/`SetOfType`/`MapType` -- it is consumed only by the clojure.spec target. So the new collection-length caps (`external_ids`/`external_references`, sighting `targets`/`observables`/`relations`/data-table, incident `categories`/`assignees`/`tactics`/`techniques`, `relation_info.actions`) are present in spec but absent from the plumatic-schema object and the generated Swagger/JSON Schema. This confirms the prior reviewer's Swagger concern at the source level. The string-length caps (`ShortString`/`MedString`/`LongString`) do propagate to both targets. Decide whether flanders should honor `:spec` on collection types in the schema target, or document the collection caps as spec-only and accept that Swagger and any plumatic-based validation will not enforce them.

- **HIGH** [correctness/cross-cutting] -- **Backward-incompatible narrowing rejects previously-valid data**: replacing unbounded `f/any-str` with bounded types is a hard reject for documents that validated before. Highest risk is `open_IOC -> MedString (2048)` -- the field's own docstring calls it "an XML blob of an openIOC indicator," and OpenIOC XML routinely exceeds 2048 chars; `SIOC` and `snort_sig` (both MedString 2048) are similar. No migration path, grace period, or source-conditional limit is present. Recommend auditing existing PROD data against each new bound and confirming `open_IOC`/`SIOC` at 2048 is intended. (See inline on `indicator.cljc`.)

- **MEDIUM** [security/correctness] -- **`relation_info.actions` cap is trivially bypassable**: the `:spec` only bounds `:actions` when it is `sequential?`; a non-sequential value (JSON object -> map, or a string) returns `true` and passes unbounded, and only the literal `:actions` key is inspected in an otherwise free-form map. As a DoS-size guard on an already-open field this is best-effort, but the cap as written does not bound `relation_info`. (See inline on `common.cljc`.)

- **MEDIUM** [quality/cross-cutting] -- **Asymmetric bounding**: the embedded `SightingDataTable` bounds `:columns`/`:rows`, but the standalone `DataTable` entity (`data_table.cljc:53-56`) leaves the structurally identical fields unbounded -- the first-class entity is the larger surface and is left open. Same pattern with `incident` `:detection_sources`, uncapped while its sibling collection fields were capped. (See inline on `incident.cljc`.)

- **MEDIUM** [tests] -- **No failure-path tests for the new bounds**: the diff touches only `src/`. None of the 12 new collection caps (nor the boundary/off-by-one, nor the 3-branch `relation_info` spec, nor the new `{:max-len}` metadata) has a test asserting rejection or at-limit acceptance. The existing generative tests only assert normally-sized documents are accepted (test.check default sizes never approach 500-10000), so "139 tests pass" does not exercise any new bound. Suggest at-N (accept) / at-N+1 (reject) tests per bound.

### PASS: Coding Style (only minor `:spec` indentation drift at `sighting.cljc:48,51` and `incident.cljc:171`; CLJS require parity is correct in all files -- no build break).

Also left 4 inline comments.

<details><summary>Non-blocking follow-ups</summary>

- `max-len` now returns `(with-meta fn {:max-len len})`; no consumer reads it in this PR, and the metadata is lost once the predicate is composed inside `cs/and` (which is how every string type uses it). Confirm the intended consumer reads it off the bare predicate.
- Centralize or document the convention for where size constants live (`common.cljc` defaults vs `sighting.cljc`-local) and add the P99.9 rationale (currently only in commit messages) as in-code comments.
- `Observable :value` is intentionally left `f/any-str` (per commit message: process_args up to ~32K); noted as deliberate, not a defect.
- `count`-based `max-len` is not guarded against a non-counted lazy seq reaching the `relation_info` spec (in-process/EDN callers only; JSON input is unaffected).

</details>

-- auto-review-team (BOT_APPROVABLE gate)

